PR #16, opened 2016-06-26
add door lock

rhooper:
This patch adds Lock support so I can later contribute lock support to HomeAssistant.


pavoni:
Now released as 0.2.11.

PR #18, opened 2016-06-28
Locks don't see to ever leave the in progress state in a timely manner.

rhooper:
This appears to be sufficient to preserve behaviour with all other devices, but make it so events get seen for locks as they happen.

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -582,6 +582,8 @@ class VeraCurtain(VeraSwitch):
 class VeraLock(VeraDevice):
     """Class to represent a doorlock"""
 
+    is_lock = True
+
     def __init__(self, json_obj, vera_controller):
         super().__init__(json_obj, vera_controller)
 
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -582,8 +582,6 @@ class VeraCurtain(VeraSwitch):
 class VeraLock(VeraDevice):
     """Class to represent a doorlock"""
 
-    is_lock = True
-
     def __init__(self, json_obj, vera_controller):
         super().__init__(json_obj, vera_controller)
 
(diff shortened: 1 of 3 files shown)


pyvera/subscribe.py
```diff
@@ -64,16 +63,21 @@ def _event_device(self, device, device_data):
         sending = comment.find('Sending') >= 0
         if sending and state == STATE_NO_JOB:
             state = STATE_JOB_WAITING_TO_START
-        if (
+        if state == STATE_JOB_IN_PROGRESS and hasattr(device, 'is_lock'):
```
pavoni: To make it similar to the code above - perhaps just set the state if it's wrong.
Also if you check the class you can avoid needing to add an attribute.
Might as well remove the error trace now you've got it working!
````python
    if state == STATE_JOB_IN_PROGRESS and isinstance(devices[0], VeraLock):
        # VeraLocks don't complete - so force state
        state = STATE_JOB_DONE
````

File: pyvera/__init__.py [hunk 2]
```diff
@@ -582,6 +582,8 @@ def set_level(self, level):
 class VeraLock(VeraDevice):
     """Class to represent a doorlock"""
 
+    is_lock = True
```
pavoni: If you check for the class - you can remove this attribute.

pyvera/subscribe.py
```diff
@@ -64,16 +63,21 @@ def _event_device(self, device, device_data):
...
                 state == STATE_JOB_WAITING_TO_START or
                 state == STATE_JOB_IN_PROGRESS or
                 state == STATE_JOB_WAITING_FOR_CALLBACK or
                 state == STATE_JOB_REQUEUE or
                 state == STATE_JOB_PENDING_DATA):
             return
-        if not (state == STATE_JOB_DONE or
-                state == STATE_NOT_PRESENT or
-                state == STATE_NO_JOB):
+        elif not (state == STATE_JOB_DONE or
```
pavoni: If you set the state - you can revert this.

pavoni:
Great work tracking this down!.

A couple of minor suggestions, but very happy to merge. 

Thanks very much for the PR

pavoni:
Thanks - merged and released.

PR #19, opened 2016-06-28
Add power consumption.

pavoni:
Add power consumption property.

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -141,7 +141,7 @@ class VeraController(object):
         else:
             devices = []
             for device in self.devices:
-                if (device.category != None and device.category != '' and 
+                if (device.category != None and device.category != '' and
                        device.category in category_filter):
                     devices.append(device)
             return devices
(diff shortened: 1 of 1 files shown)


snizzleorg:
how this should show up in home-assistant? as an additional sensor?

pavoni:
@snizzleorg Afraid I'm not in the same location as my Vera - won't be there until next week, so can't test.

This PR will add the power attribute to existing HA switches. You should see it as an additional attribute. https://github.com/home-assistant/home-assistant/pull/2490

You can break it out as a dedicated sensor if you want to using a sensor template.

I don't know if there are other devices that also support the power attribute.

snizzleorg:
hm. for mine it doesn't yet show up as additional property. This is with hass 0.23.1...not sure if the commit is already included in hass 0.23.1

pavoni:
The PR isn't merged! You can try my test branch if you run a dev build.

Can't merge until it's been tested!

snizzleorg:
shit. I can't test that easy since im running a docker image.

PR #21, opened 2016-07-01
couldn't import VeraLock in subclass

rhooper:
This ughly hack should addres `16-07-01 13:19:29 pyvera.subscribe: Vera thread exception name 'VeraLock' is not defined`

You might want to just hotfix 0.2.12 instead, although I'm happy to send a MR to HA to bump the requirements to 0.2.13.


pavoni:
Would be nice to find a cleaner fix - but sounds urgent to will merge and release.

pavoni:
Released as 0.2.13.

PR #22, opened 2016-07-01
Address flake8 and pylint for code.

rhooper:
This PR cleans up style to pass flake8 and pylint, but does not fix the VeraLock bug.
It does not remove fix redundancy, duplication, and repetition of code.


pavoni:
Thanks for the cleanup.

Issue #24, opened 2016-07-03
VeraLite dimmers and locks


amccardie:
For whatever reason, my VeraLite (UI 5, fw 1.5.672) returns different category names for dimmers and door locks than pyvera expects.

Specifically, instead of "Dimmable Switch", VeraLite returns "Dimmable Light" and instead of "Doorlock" it returns "Door lock".

Home Assistant can't find either my dimmers or my lock unless I change the strings in both pyvera/__init__.py  and homeassistant/components/vera.py. 

Let me know if you need more information or if I can help out.

pavoni:
You could upgrade your Vera. I know there were issues - but mine works pretty well on the latest UI-7.

However if you don't want to upgrade, happy to add another set of device strings. Is it just dimmers and locks that don't work?

Either put together a PR if you're a tech guy, or make a list comparing the ones we use in the code - and I'll put something together.

On vacation this week - so likely to take a couple of weeks.

amccardie:
In my case, yes, it's just dimmers and locks. I've been avoiding UI7 mainly because UI5 has been very stable for me. If upgrading to UI7 will fix my issue though, I may take the plunge.

I can patch the files pretty quickly when HA is updated, so there's no urgency at all. Also, if I get a chance, I'll try to set up a pull request. I haven't worked with GitHub (or git) before, so it will be an adventure.

pavoni:
Included in https://github.com/pavoni/pyvera/pull/27 and also in the dev branch of HA.

PR #25, opened 2016-07-09
Add detection support for UI5 dimmers and locks (issue #24)

amccardie:
These changes allow detection of dimmers and locks from UI5 VeraLite and possibly other UI5 Vera controllers. No modifications to Home-Assistant's Vera component are required.

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -152,7 +152,7 @@ class VeraController(object):
                   item.get('deviceInfo').get('categoryName') ==
                   'Door lock'):
                 doorlock = VeraLock(item, self)
-                doorlock.category = "Doorlock"
+                doorlock.category = 'Doorlock'
                 self.devices.append(doorlock)
             else:
                 self.devices.append(VeraDevice(item, self))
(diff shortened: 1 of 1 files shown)


pyvera/__init__.py
```diff
@@ -119,6 +119,12 @@ def get_devices(self, category_filter=''):
                 self.devices.append(VeraDimmer(item, self))
             elif (item.get('deviceInfo') and
                   item.get('deviceInfo').get('categoryName') ==
+                  'Dimmable Light'):        
+                dimmer = VeraDimmer(item, self)
+                dimmer.category = "Dimmable Switch"
```
pavoni: Rather that changing the category - I think it might be better to get the same effect by overriding `is_dimmable` in VeraDimmer:-

````python
    @property
    def is_dimmable(self):
        """Device is dimmable."""
        return True
````
That way - once we've decided the object is a Dimmer - other checks will be consistent.

For purity also probably better to refactor the base `VeraDevice` class slightly:-
````python
    @property
    def is_dimmable(self):
        """Default device is not dimmable."""
        return False
````

Longer term I think this would be a better pattern for the others objects too - I will refactor along these lines

pavoni: I can see this requires more thought with HA - will merge for now - and look to refactor myself later.

pavoni:
Thanks for the contribution.

I'm away from my Vera - so will merge and release a version. Be grateful if you could check the release with HA before releasing there.

pavoni:
Now released as 0.2.14

Issue #20, opened 2016-06-28
thermostat support


snizzleorg:
... would be very nice as well. I have a couple of danfoss z-wave thermostats and of course would love to be able to control them via home-assistant.

pavoni:
Certainly support adding this - but I don't have any connected thermostats. 

Would welcome a contribution, and happy to help!

snizzleorg:
thanks. I'm not really a programmer. I can work myself through some python code but i have never actually done any real programming. 

Of course I can test the code and give feedback. I guess remote access to my setup could also be provided.

pavoni:
@snizzleorg Thanks for the offer. I think adding a thermostat will take some experimenting - so would need to be a developer with local access. Hopefully we'll find someone!

robjohnson189:
I started working on adding thermostat support, It doesn't look like it is going to take too long to get this added. @snizzleorg what kind of thermostat do you have? I'll be developing this to work against a Radio Thermostat CT80. When I paired my thermostat with vera, it shows up as 3 different devices in the Vera Devices page but it seems that I can only really control the thermostat using 1 of those 3...

snizzleorg:
I have danfoss living connect (http://www.pepper1.net/zwavedb/device/426)

not sure which firmware version though. Let me know if I can help somehow...

robjohnson189:
@snizzleorg  I've got this working on my end. If you would like to help me test you can install and run my fork of pyvera (https://github.com/robjohnson189/pyvera) and home-assistant (https://github.com/robjohnson189/home-assistant). Let me know how it works. It seems to be working good for me but my thermostat has quite a few more features than yours.

By default home-assistant will load any thermostats as climate components, if you want them loaded into home-assistant as a thermostat you can add the config "climate: thermostat" to your vera config.

@pavoni After a little more testing I'll create a pr

snizzleorg:
@robjohnson189 Thanks!
 I'm running a docker image so it's not that easy for me to test. I'll try if I get your branch running in a docker image - hopefully I'll get to it on the weekend.

pavoni:
Now added - so closing.

snizzleorg:
@pavoni I still don't see my thermostats. But I was wondering if this is maybe because they show up as heater in vera? 
They are Radiator Thermostats (danfoss living connect)
[...]

PR #28, opened 2016-08-29
thermostat support #20

robjohnson189:
Added vera thermostat support.

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -158,6 +158,10 @@ class VeraController(object):
                   item.get('deviceInfo').get('categoryName') ==
                   'Door lock'):
                 self.devices.append(VeraLock(item, self))
+            elif (item.get('deviceInfo') and
+                  item.get('deviceInfo').get('categoryName') ==
+                  'Thermostat'):
+                self.devices.append(VeraThermostat(item, self))
             else:
                 self.devices.append(VeraDevice(item, self))
 
@@ -272,6 +276,16 @@ class VeraDevice(object):
             else:
                 self.name = 'Vera Device ' + str(self.device_id)
 
+    def get_payload(self, service_name, name, value):
+        """the http payload for setting a variable"""
+        return {
+            'id': 'lu_action',
+            'output_format': 'json',
+            'DeviceNum': self.device_id,
+            'serviceId': service_name,
+            'action': 'Set' + name,
+            'new' + name + 'Value': value}
+
     def set_value(self, name, value):
         """Set a variable on the vera device.
 
@@ -293,13 +307,7 @@ class VeraDevice(object):
                         'action': 'Set' + name,
                         'newLoadlevelTarget': value}
                 else:
-                    payload = {
-                        'id': 'lu_action',
-                        'output_format': 'json',
-                        'DeviceNum': self.device_id,
-                        'serviceId': service_name,
-                        'action': 'Set' + name,
-                        'new' + name + 'Value': value}
+                    payload = self.get_payload(service_name, name, value)
                 request_url = self.vera_controller.base_url + "/data_request"
                 requests.get(request_url, params=payload)
                 item['value'] = value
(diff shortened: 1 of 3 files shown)


pyvera/__init__.py
```diff
@@ -293,13 +307,7 @@ def set_value(self, name, value):
                         'action': 'Set' + name,
                         'newLoadlevelTarget': value}
                 else:
-                    payload = {
```
pavoni: I think you've moved half the logic to a new function here, which slightly obscures the issue over handling handling the `newLoadlevelTarget` special case. 

The existing code isn't great - certainly worth refactoring and perhaps handling `newLoadlevelTarget` issue as a special case without repeating the whole dictionary construction.

Happy for you to complete the refactor - or just put back (and I'll do it another time) as you wish.

pyvera/__init__.py
```diff
@@ -645,3 +653,91 @@ def is_locked(self, refresh=False):
...
+    """Class to represent a thermostat."""
+    def get_payload(self, service_name, name, value):
+        return {
+            'id': 'lu_action',
+            'output_format': 'json',
+            'DeviceNum': self.device_id,
+            'serviceId': service_name,
+            'action': 'Set' + name,
+            'New' + name: value}
+
```
pavoni: This method looks quite generic - how are you using it? 

Don't think I can see it being used elsewhere in the class.

It looks very generic so my thought would be either to decide it's generally useful and add it into the base class - or drop. What do you think?

pavoni:
Couple of small comments above.

Thanks very much for the contribution - particularly as it looks like it require some detective work (`mode` vs `ModeTarget` etc)!

Happy to merge and release when we've finished the discussions. I think you're an HA user - so guess you have updated the HA wrapper code too? We should be able to get it into the next release.

Thanks again!

pyvera/__init__.py
robjohnson189: This method is in the base class (see line 279), no other class overrides it, any entity that isn't a thermostat uses the base class implementation.

pyvera/__init__.py
robjohnson189: I would have moved the LoadLevelTarget logic to a decedent class method override of get_payload but I don't know which entity that should be a part of. If you know which entity needs the parameter constructed that way I'd be happy to move it to the get_payload method of that entity.

The reason I moved it to a method is because Thermostat required a different dictionary than either of the original dictionaries provided (the name=LoadLevelTarget or Else case prior to my change).

I'll clean this up a bit so we can re-use creating the majority of the dictionary, while only adding the last parameter to the dictionary but allowing the sub class or base class to create that parameter.

After my change should work as follows:
1. create dictionary with id, output_format, DeviceNum, serviceId and action populated.
2. add last parameter to dictionary
    2a. (if themostat) New<name> = x (provided by VeraThermostat class)
    2b. (if name != LoadLevelTarget) new<name>Value = x (provided by base class)
    2c. (if name == LoadLevelTarget) newLoadLevelTarget = x (provided by if statement, unless you can tell me which class needs this)

Make sense?

pyvera/__init__.py
robjohnson189: this method is currently class in line 310, if the decedent class doesn't override this (every class except VeraThermostat) then the base class provides the value.

robjohnson189:
Yes I have updated the HA wrapper code, but want to get this merged and released before I create the pull request for HA. I'll make sure to add a comment with your username on the HA Pull request so it comes to your attention when I create it.

pyvera/__init__.py
pavoni: Got it - I missed that you needed to override.

pyvera/__init__.py
pavoni: Yes, let's clean this up. Of our current devices `LoadLevelTarget` is used by `VeraDimmer` and `VeraCurtain`.

I'll do some digging to see if I can find what else it might be used for.

pavoni: Some docs here http://wiki.micasaverde.com/index.php/Luup_UPnP_Variables_and_Actions although just claims to be the dimmer!

Still reading...

pavoni: The `new{name}Value` isn't much of a standard from what I can see in the docs - although it does work for `Target` and `Armed`, both of which we use.

`new{name}` is perhaps used more after your additions.

I think I'd be tempted to have a list of `exceptions` and handle both in the base class. 

If there are other examples like the `curtain` where the same variable is re-used, this will handle it without further work.

But don't have strong feelings (except that Vera should have stuck to one standard!).

robjohnson189:
How about that?

robjohnson189:
Now a version that works with Dimmer Switches

pyvera/__init__.py
```diff
@@ -488,6 +478,11 @@ def is_switched_on(self, refresh=False):
 class VeraDimmer(VeraSwitch):
     """Class to add dimmer functionality."""
 
+    def get_payload_parameter_name(self, name):
+        if name == "LoadLevelTarget": # LoadLevel to Loadlevel, api is case sensitive
+            return "newLoadlevelTarget"
+        return "New" + name
```
pavoni: I think this should also be `new` rather than `New`.

pavoni: or better 
````
return super(VeraDimmer, self).get_payload_parameter_name()
````

File: pyvera/__init__.py [hunk 4]
```diff
@@ -576,6 +579,11 @@ def is_switched_on(self, refresh=False):
 class VeraCurtain(VeraSwitch):
     """Class to add curtains functionality."""
 
+    def get_payload_parameter_name(self, name):
+        if name == "LoadLevelTarget": # LoadLevel to Loadlevel, api is case sensitive
+            return "newLoadlevelTarget"
+        return "New" + name
```
pavoni: `return super(VeraDimmer, self).get_payload_parameter_name(`

pavoni: `return super(VeraDimmer, self).get_payload_parameter_name()`

pavoni:
Looks good, thanks for doing this. A couple of comments.

pavoni:
Thanks - now released as 0.2.16.

PR #34, opened 2016-11-13
Add basic support for siren on/off switch device

roberttuttle:


pavoni:
Hi Robert, thanks for the contribution.

I see you've created a new class for the Siren, although at this point it doesn't have any additional behaviour over the switch. I did this for the Sensor because one of the main users of the library needed to distinguish the sensor from other devices, but generally I've only added a class where there was some functionality.

If you don't need an extra class you can just create a `VeraSwitch`.

roberttuttle:
@pavoni ... thanks for the feedback. Agreed, premature PR. I am digging into the details of the Siren classes that also allow volume and sound type control versus just an on/off switch control (along with being armed/disarmed). I'll close this and reopen when ready.

PR #35, opened 2016-11-20
Add timeout to all requests calls

sdague:
I am seeing a situation where the pyvera component just stops getting
new info in home-assistant, no errors logged. In looking through the
requests documentation it looks like if requests calls don't have a
timeout, then requests can hang forever -
http://docs.python-requests.org/en/master/user/quickstart/#timeouts

This adds a timeout to all calls, but most importantly is the poll
loop.

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -14,6 +14,8 @@ __author__ = 'jamespcole'
 SUBSCRIPTION_WAIT = 30
 # Min time to wait for event in miliseconds
 SUBSCRIPTION_MIN_WAIT = 200
+# Timeout for requests calls, as vera sometimes just sits on sockets.
+TIMEOUT = SUBSCRIPTION_WAIT
 
 
 _VERA_CONTROLLER = None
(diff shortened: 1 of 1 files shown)


pavoni:
Makes sense, and code looks good.

If the timeout happens - were you able to check we cope OK. Pretty sure the poll loop will be fine (I think it's just a request exception which we already catch and handle) - which is the most important case.

sdague:
I didn't check the failure case yet, I'm running this patch on my HA setup now, kind of hope that at least when the lock up happens again there will be a clear enough exception that it can be addressed. It's a pretty rare event (once every couple of weeks) but it does mean a whole chunk of my home automation goes dark.

pavoni:
I have occasionally seen something similar 

Perhaps you could test by temporarily change the poll timeout to a much lower number and check it copes. That's the key one (you may have already done this in testing). 

If that's OK I'm happy to merge & release.

sdague:
Ok. I set the timeout to 5 seconds, and it all seems to fail gracefully. This is what my logs are looking like after that.

16-11-21 05:35:33 INFO (Vera Poll Thread) [pyvera.subscribe] Could not contact Vera - will retry in 10s
16-11-21 05:35:44 INFO (Vera Poll Thread) [pyvera.subscribe] Poll returned
16-11-21 05:35:47 INFO (Vera Poll Thread) [pyvera.subscribe] Poll returned
16-11-21 05:35:53 INFO (Vera Poll Thread) [pyvera.subscribe] Poll returned

sdague:
Thanks! Let me know once a release is out and I'll get the HA req patch in to use this.

pavoni:
0.2.21 now uploaded. Thanks for the contribution.

PR #36, opened 2016-12-03
Thermostat improvements

jnewland:
This PR makes two changes to more completely support thermostats w/ multiple modes and setpoints like my TBZ48.

2e6e2a0 extracts cool, heat, and auto setpoints from `AllSetpoints` as documented [here](http://wiki.micasaverde.com/index.php/Luup_UPnP_Variables_and_Actions#TemperatureSetpoint1).

f5ef281 supports calling the `SetModeTarget` if your device doesn't have a `ModeTarget` state variable. For whatever reason, my device only reports a `ModeStatus` state variable, which prevented `set_value('ModeTarget', value)` from working because of [this check](https://github.com/pavoni/pyvera/blob/24f7dd3d615266172e2a48aa24b9b308272cab21/pyvera/__init__.py#L297).

I've tested these changes against my device like so:

```python
import pyvera, json

url = "http://IP:3480/"
controller = pyvera.VeraController(url)
thermostats = controller.get_devices('Thermostat')

device = thermostats[0]

print(device.name)
print(device.get_current_goal_temperature())
print(device.get_current_temperature())
print(device.get_current_heat_setpoint())
print(device.get_current_cool_setpoint())
print(device.get_hvac_mode())
device.turn_heat_on()
print(device.get_hvac_mode())
device.turn_auto_on()
print(device.get_hvac_mode())
```
```bash
root@077d603274cd:/repo# python test.py 
great room
65
68
65.0
73.0
AutoChangeOver
HeatOn
AutoChangeOver
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,0 @@
-FROM python:3.5
-
-VOLUME /repo
-
-RUN pip install \
-  requests \
-  pylint
-
-WORKDIR /repo
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM python:3.5
+
+VOLUME /repo
+
+RUN pip install \
+  requests \
+  pylint
+
+WORKDIR /repo
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -288,6 +288,20 @@ class VeraDevice(object):
         """the http payload for setting a variable"""
         return 'new' + name + 'Value'
 
+    def lu_action(self, action, service, payload):
+        """Perfom a lu_action."""
+        request_payload = {
+            'id': 'lu_action',
+            'output_format': 'json',
+            'DeviceNum': self.device_id,
+            'serviceId': 'urn:upnp-org:serviceId:' + service,
+            'action': action
+        }
+        request_payload.update(payload)
+        request_url = self.vera_controller.base_url + "/data_request"
+        requests.get(request_url, timeout=TIMEOUT, params=request_payload)
+        return request_payload
+
     def set_value(self, name, value):
         """Set a variable on the vera device.
 
@@ -706,7 +720,9 @@ class VeraThermostat(VeraDevice):
 
     def set_hvac_mode(self, mode):
         """Set the hvac mode"""
-        self.set_value('ModeTarget', mode)
+        self.lu_action('SetModeTarget', 'HVAC_UserOperatingMode1', {
+            'NewModeTarget': mode
+        })
         self.set_cache_value('mode', mode)
 
     def get_hvac_mode(self, refresh=False):
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -288,19 +288,19 @@ class VeraDevice(object):
         """the http payload for setting a variable"""
         return 'new' + name + 'Value'
 
-    def lu_action(self, action, service, payload):
-        """Perfom a lu_action."""
+    def action(self, request_id, action, service, payload):
+        """Perfom a data_request action."""
         request_payload = {
-            'id': 'lu_action',
+            'id': request_id,
             'output_format': 'json',
             'DeviceNum': self.device_id,
-            'serviceId': 'urn:upnp-org:serviceId:' + service,
+            'serviceId': service,
             'action': action
         }
         request_payload.update(payload)
         request_url = self.vera_controller.base_url + "/data_request"
-        requests.get(request_url, timeout=TIMEOUT, params=request_payload)
-        return request_payload
+        return requests.get(request_url, timeout=TIMEOUT,
+                            params=request_payload)
 
     def set_value(self, name, value):
         """Set a variable on the vera device.
@@ -310,16 +310,10 @@ class VeraDevice(object):
         for item in self.json_state.get('states'):
             if item.get('variable') == name:
                 service_name = item.get('service')
-                payload = {
-                    'id': 'lu_action',
-                    'output_format': 'json',
-                    'DeviceNum': self.device_id,
-                    'serviceId': service_name,
-                    'action': 'Set' + name
-                }
-                payload.update({self.get_payload_parameter_name(name): value})
-                request_url = self.vera_controller.base_url + "/data_request"
-                requests.get(request_url, timeout=TIMEOUT, params=payload)
+
+                self.action('lu_action', 'Set' + name, service_name, {
+                    self.get_payload_parameter_name(name): value
+                })
                 item['value'] = value
 
     def call_service(self, service_name, action):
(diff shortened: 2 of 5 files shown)


pavoni:
@robjohnson189 as the thermostat expert do you have any thoughts on this? (I don't have a vera thermostat)

My only thought is that with the addition of `lu_action` we now have 3 versions of basically the same code, so we should refactor (I should have done this earlier when I added the second). Would be great if you were willing to do this - otherwise I can take a look when I get a chance.

jnewland:
> Would be great if you were willing to do this - otherwise I can take a look when I get a chance.

I'm definitely willing to take a stab at this, and should have some time tomorrow morning. I'll  mention you when I have something worthy of review!

pyvera/__init__.py
```diff
@@ -284,10 +284,25 @@ def __init__(self, json_obj, vera_controller):
             else:
                 self.name = 'Vera Device ' + str(self.device_id)
 
-    def get_payload_parameter_name(self, name):
+    @staticmethod
```
pavoni: This is overridden by some derived classes, some of which then call the base class version, so I don't think it's right to make this `@staticmethod`, although your python may be better than mine...

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.5
```
pavoni: Could you remove the docker related files?

script/bootstrap
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+docker build -t pyvera .
```
pavoni: Please see above.

script/console
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
pavoni: Please see above

script/lint
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
pavoni: please see above

Dockerfile
jnewland: Sure! Would you be up for accept another, separate pull request that adds them to a `docker` directory along with a section in the readme that documents how how to run some TBD tests for this library?

pyvera/__init__.py
jnewland: My python probably isn't better than yours: `pylint` was complaining about this method since it doesn't use `self`. I'll back out the change.

pavoni:
At a quick glance his looks great - thanks for taking the time to do this.

Do you want me to review / test / merge at this point - or were there more changes you wanted to make?

jnewland:
> Do you want me to review / test / merge at this point - or were there more changes you wanted to make?

I have some local changes that I'd like to push up after doing a bit more research into how my thermostats are behaving in the real world. Unfortunately, it looks like the `heat` and `cool` values in `sdata` for my thermostat don't always reflect reality. Sorry for the delay!

pavoni:
No problem - and no rush. Just didn't want to be holding you up.

pavoni:
Happy new year!. Just checking how things are looking. Let me know if I can help.

jnewland:
@pavoni hey, happy new year to you too! Thanks for checking in. 😄 There's a bunch going on in this PR:

* f5ef2819e66d3595ea6b393253e00abaa0335106 has been working well for me in practice, and is an important improvement.
* I've just pushed up c35edf0, which completes the refactoring you mentioned in https://github.com/pavoni/pyvera/pull/36#issuecomment-264673719. I haven't yet tested it in practice, but will do so immediately following this PR.
* 2e6e2a0 was a neat idea to workaround some strange Vera behavior in Auto mode, but I've since reverted it since there are a handful of negative tradeoffs. A Vera support rep clarified what's going on here:

> Auto Mode will read the values of Heat and Cool and make the average. For example if you set Heat to 80 and Cool to 70, Auto will read those values and bring the temperature to 75.

The only way for us to extract independent setpoints from a Thermostat in Auto mode would be to preform a full status request on an interval, which doesn't seem like a thing that makes sense to do just yet. This followup from the Vera support rep seems to indicate that they don't consider interacting with a Thermostat in Auto mode to be a thing they'd like to improve:

> Auto is not a mode which you should set up manually, would be better if you could set your Heat and Cool setpoints and then leave Auto to work on it’s own making the average and running on that value.

So, effectively, this PR now has this function: it fixes calling the `SetModeTarget` if your device doesn't have a `ModeTarget` state variable, refactors the `lu_action` code into one `data_request` method, and fixes a lint complaint. Mind taking another pass at code review? I'm booting up home-assistant with this branch right now to confirm practical functionality.

pyvera/__init__.py
jnewland: I looked into this just now, and do believe that this is a static method after a read of [this comment](http://stackoverflow.com/questions/136097/what-is-the-difference-between-staticmethod-and-classmethod-in-python/1669524#1669524):

> With staticmethods, neither self (the object instance) nor  cls (the class) is implicitly passed as the first argument. They behave like plain functions except that you can call them from an instance or the class. ... Staticmethods are used to group functions which have some logical connection with a class to the class.

Does this make sense to you?

jnewland:
>  I'm booting up home-assistant with this branch right now to confirm practical functionality.

Welp, I broke something 😬. Turning off and on lights is broken in home-assistant with this branch. I'll ping you again when I've confirmed things are at 💯.

pyvera/__init__.py
pavoni: I think the problem is that you can't override static methods in inherited classes.

You probably want a class method.  Not a corner of Python I've explored - but his piece looked helpful.
https://julien.danjou.info/blog/2013/guide-python-static-class-abstract-methods

pavoni: This is clearer https://rapd.wordpress.com/2008/07/02/python-staticmethod-vs-classmethod/

jnewland: I'm definitely a Python noob, so rather than make any concrete changes in an area that seems vague (the first link says an *advantage* of static methods is that they allow the developer to  override them in subclasses, and the second says their definition is immutable ❓ ), I've backed out this change in 4e9d4e7 and added a comment to disable the lint error it generates instead.

jnewland:
> Welp, I broke something 😬. Turning off and on lights is broken in home-assistant with this branch. I'll ping you again when I've confirmed things are at 💯.

af49cac appears that it was the fix here. I'm running this branch locally and things are working as expected. I've confirmed the functionality of lights, as well as setting a thermostat mode and setpoint.

pavoni:
This look great. I'll merge and test locally, and then release assuming all looks good. 

Thanks very much for the contribution.

pavoni:
Works fine on my devices - so now released as `0.2.22`. Will leave the HA update to you (since it's your work)! Thanks again.

robjohnson189:
Works fine on my switches and thermostat as well.

pavoni:
Thanks for testing Rob.

PR #37, opened 2017-01-07
Support for controlling RFXtrx devices in Vera.

angoyd:
Since RFX devices don´t expose variable Target with correct service they
cant be controlled in the current implemenation, I have changes to its checking if its an RFX devices and changes the servicename to the correct one.


pavoni:
Thanks for this. What you've done makes sense - but inspired me to doa rafactor I've been thinking about for a while which should solve your issue more elegantly - and clean up some other problems.

Could you take a look at https://github.com/pavoni/pyvera/pull/38 and see if this works for you?

angoyd:
#38 works just like supposed to 👍 a cleaner solution to the problem.

PR #38, opened 2017-01-08
Revise set logic

pavoni:
This PR changes (most) of the classes to explicitly call the Vera service they need - rather than having the `set_value` code have a `guess` about which service to call.

Vera isn't 100% consistent about the naming convention when setting variables - so we've had some logic for a while that allowed the convention to be over-ridden.

Recent additions (window coverings, thermostats) have needed to be explicit about what services they wanted to call.

Plus this PR https://github.com/pavoni/pyvera/pull/37 felt like the final nail in the coffin since even for a switch the `set_value` service selection wasn't correct.

I've tested the code on my own devices - but I don't have a lock a thermostat - so would be grateful for some help testing on these devices.

Finally I've left one thermostat service (`CurrentSetPoint`) calling set_level - because I'm not 100% sure which service to use (and I didn't want to break the thermostat code). Looks like vera has services for both Heat and Cool. I don't know which one we are calling at present (see http://wiki.micasaverde.com/index.php/Luup_Devices#SetModeTarget).

I've also tidied up the ugly if statement a little.

Grateful for any help from @jnewland @robjohnson189 @rhooper and @angoyd in making sure this works OK.

Once we no longer call `set_value` I'll tidy up the code some more.

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -66,16 +66,14 @@ class VeraController(object):
         self.categories = {}
         self.device_id_map = {}
 
-    def data_request(self, **kwargs):
+    def data_request(self, payload, timeout=TIMEOUT):
         """Perform a data_request and return the result."""
-        payload = dict(kwargs)
-        timeout = payload.pop('timeout', TIMEOUT)
         request_url = self.base_url + "/data_request"
         return requests.get(request_url, timeout=timeout, params=payload)
 
     def get_simple_devices_info(self):
         """Get basic device info from Vera."""
-        j = self.data_request(id='sdata').json()
+        j = self.data_request({'id': 'sdata'}).json()
 
         if j.get('temperature'):
             self.temperature_units = j.get('temperature')
@@ -104,7 +102,7 @@ class VeraController(object):
         # all the info e need
         self.get_simple_devices_info()
 
-        j = self.data_request(id='status', output_format='json').json()
+        j = self.data_request({'id': 'status', 'output_format': 'json'}).json()
 
         self.devices = []
         items = j.get('devices')
@@ -161,7 +159,7 @@ class VeraController(object):
 
     def refresh_data(self):
         """Refresh data from Vera device."""
-        j = self.data_request(id='sdata').json()
+        j = self.data_request({'id': 'sdata'}).json()
 
         self.temperature_units = j.get('temperature', 'C')
         self.model = j.get('model')
@@ -189,7 +187,7 @@ class VeraController(object):
         # to get all the info e need
         self.get_simple_devices_info()
 
-        j = self.data_request(id='status', output_format='json').json()
+        j = self.data_request({'id': 'status', 'output_format': 'json'}).json()
 
         service_map = {}
 
@@ -216,9 +214,8 @@ class VeraController(object):
         # double the timeout here so requests doesn't timeout before vera
         payload.update({
             'id': 'lu_sdata',
-            'timeout': TIMEOUT * 2
         })
-        result = self.data_request(**payload).json()
+        result = self.data_request(payload, TIMEOUT*2).json()
         device_data = result.get('devices')
         timestamp = {
             'loadtime': result.get('loadtime'),
@@ -302,15 +299,15 @@ class VeraDevice(object):  # pylint: disable=R0904
         """the http payload for setting a variable"""
         return 'new' + name + 'Value'
 
-    def data_request(self, **kwargs):
-        """Perfom a data_request for this device."""
+    def vera_request(self, **kwargs):
+        """Perfom a vera_request for this device."""
         request_payload = {
             'output_format': 'json',
             'DeviceNum': self.device_id,
         }
         request_payload.update(kwargs)
 
-        return self.vera_controller.data_request(**request_payload)
+        return self.vera_controller.data_request(request_payload)
 
     def set_value(self, name, value):
         """Set a variable on the vera device.
@@ -328,8 +325,8 @@ class VeraDevice(object):  # pylint: disable=R0904
                     'serviceId': service_id,
                     self.get_payload_parameter_name(name): value
                 }
-                result = self.data_request(**payload)
-                LOG.debug("Result of data_request with payload %s: %s",
+                result = self.vera_request(**payload)
+                LOG.debug("Result of vera_request with payload %s: %s",
                           payload,
                           result.text)
 
@@ -346,8 +343,8 @@ class VeraDevice(object):  # pylint: disable=R0904
             'serviceId': service_id,
             parameter_name: value
         }
-        result = self.data_request(**payload)
-        LOG.debug("Result of data_request with payload %s: %s", payload,
+        result = self.vera_request(**payload)
+        LOG.debug("Result of vera_request with payload %s: %s", payload,
                   result.text)
 
     def call_service(self, service_id, action):
@@ -355,7 +352,7 @@ class VeraDevice(object):  # pylint: disable=R0904
 
         This will call the Vera api to change device state.
         """
-        return self.data_request(id='action', serviceId=service_id,
+        return self.vera_request(id='action', serviceId=service_id,
                                  action=action)
 
     def set_cache_value(self, name, value):
@@ -403,7 +400,7 @@ class VeraDevice(object):  # pylint: disable=R0904
         for item in self.json_state.get('states'):
             if item.get('variable') == name:
                 service_id = item.get('service')
-                result = self.data_request(**{
+                result = self.vera_request(**{
                     'id': 'variableget',
                     'output_format': 'json',
                     'DeviceNum': self.device_id,
@@ -419,7 +416,7 @@ class VeraDevice(object):  # pylint: disable=R0904
 
         Only needed if you're not using subscriptions.
         """
-        j = self.data_request(id='sdata', output_format='json').json()
+        j = self.vera_request(id='sdata', output_format='json').json()
         devices = j.get('devices')
         for device_data in devices:
             if device_data.get('id') == self.device_id:
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -22,6 +22,7 @@ _VERA_CONTROLLER = None
 
 LOG = logging.getLogger(__name__)
 
+
 def init_controller(url):
     """Initialize a controller.
 
@@ -110,67 +111,41 @@ class VeraController(object):
 
         for item in items:
             item['deviceInfo'] = self.device_id_map.get(item.get('id'))
-            if (item.get('deviceInfo') and
-                    item.get('deviceInfo').get('categoryName') ==
-                    'Switch'):
-                self.devices.append(VeraSwitch(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'On/Off Switch'):
-                self.devices.append(VeraSwitch(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Dimmable Switch'):
-                self.devices.append(VeraDimmer(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Dimmable Light'):
-                self.devices.append(VeraDimmer(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Temperature Sensor'):
-                self.devices.append(VeraSensor(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Humidity Sensor'):
-                self.devices.append(VeraSensor(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Light Sensor'):
-                self.devices.append(VeraSensor(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Sensor'):
-                sensor = VeraBinarySensor(item, self)
-                self.devices.append(sensor)
-                if sensor.is_armable:
-                    armable = VeraArmableDevice(item, self)
-                    armable.category = 'Armable Sensor'
-                    self.devices.append(armable)
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Light Sensor'):
-                self.devices.append(VeraSensor(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Window Covering'):
-                self.devices.append(VeraCurtain(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Window covering'):
-                self.devices.append(VeraCurtain(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Doorlock'):
-                self.devices.append(VeraLock(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Door lock'):
-                self.devices.append(VeraLock(item, self))
-            elif (item.get('deviceInfo') and
-                  item.get('deviceInfo').get('categoryName') ==
-                  'Thermostat'):
-                self.devices.append(VeraThermostat(item, self))
+            if item.get('deviceInfo'):
+                device_category = item.get('deviceInfo').get('categoryName')
+                if device_category == 'Switch':
+                    self.devices.append(VeraSwitch(item, self))
+                elif device_category == 'On/Off Switch':
+                    self.devices.append(VeraSwitch(item, self))
+                elif device_category == 'Dimmable Switch':
+                    self.devices.append(VeraDimmer(item, self))
+                elif device_category == 'Dimmable Light':
+                    self.devices.append(VeraDimmer(item, self))
+                elif device_category == 'Temperature Sensor':
+                    self.devices.append(VeraSensor(item, self))
+                elif device_category == 'Humidity Sensor':
+                    self.devices.append(VeraSensor(item, self))
+                elif device_category == 'Light Sensor':
+                    self.devices.append(VeraSensor(item, self))
+                elif device_category == 'Sensor':
+                    sensor = VeraBinarySensor(item, self)
+                    self.devices.append(sensor)
+                    if sensor.is_armable:
+                        armable = VeraArmableDevice(item, self)
+                        armable.category = 'Armable Sensor'
+                        self.devices.append(armable)
+                elif device_category == 'Light Sensor':
+                    self.devices.append(VeraSensor(item, self))
+                elif device_category == 'Window Covering':
+                    self.devices.append(VeraCurtain(item, self))
+                elif device_category == 'Window covering':
+                    self.devices.append(VeraCurtain(item, self))
+                elif device_category == 'Doorlock':
+                    self.devices.append(VeraLock(item, self))
+                elif device_category == 'Door lock':
+                    self.devices.append(VeraLock(item, self))
+                elif device_category == 'Thermostat':
+                    self.devices.append(VeraThermostat(item, self))
             else:
                 self.devices.append(VeraDevice(item, self))
 
@@ -264,8 +239,8 @@ class VeraController(object):
         self.subscription_registry.register(device, callback)
 
 
-class VeraDevice(object): # pylint: disable=R0904
-    """Class to represent each vera device."""
+class VeraDevice(object):  # pylint: disable=R0904
+    """ Class to represent each vera device."""
 
     def __init__(self, json_obj, vera_controller):
         """Setup a Vera device."""
@@ -345,7 +320,8 @@ class VeraDevice(object): # pylint: disable=R0904
         for item in self.json_state.get('states'):
             if item.get('variable') == name:
                 service_id = item.get('service')
-                LOG.debug("Set variable: %s using service_id: %s", name, service_id)
+                LOG.debug("Set variable: %s using service_id: %s", name,
+                          service_id)
                 payload = {
                     'id': 'lu_action',
                     'action': 'Set' + name,
@@ -353,7 +329,8 @@ class VeraDevice(object): # pylint: disable=R0904
                     self.get_payload_parameter_name(name): value
                 }
                 result = self.data_request(**payload)
-                LOG.debug("Result of data_request with payload %s: %s", payload,
+                LOG.debug("Result of data_request with payload %s: %s",
+                          payload,
                           result.text)
 
                 item['value'] = value
@@ -535,7 +512,7 @@ class VeraSwitch(VeraDevice):
             self.switch_service,
             'Target',
             'newTargetValue',
-             state)
+            state)
         self.set_cache_value('Status', state)
 
     def switch_on(self):
@@ -560,7 +537,6 @@ class VeraSwitch(VeraDevice):
 
 class VeraDimmer(VeraSwitch):
     """Class to add dimmer functionality."""
-
 
     def switch_on(self):
         """Turn the dimmer on."""
@@ -612,7 +588,7 @@ class VeraDimmer(VeraSwitch):
             self.dimmer_service,
             'LoadLevelTarget',
             'newLoadlevelTarget',
-             percent)
+            percent)
         self.set_cache_value('level', percent)
 
 
@@ -625,7 +601,7 @@ class VeraArmableDevice(VeraSwitch):
             self.security_sensor_service,
             'Armed',
             'newArmedValue',
-             state)
+            state)
         self.set_cache_value('Armed', state)
 
     def switch_on(self):
@@ -719,9 +695,10 @@ class VeraCurtain(VeraSwitch):
             self.dimmer_service,
             'LoadLevelTarget',
             'newLoadlevelTarget',
-             level)
+            level)
 
         self.set_cache_value('level', level)
+
 
 class VeraLock(VeraDevice):
     """Class to represent a door lock."""
@@ -732,7 +709,7 @@ class VeraLock(VeraDevice):
             self.lock_service,
             'Target',
             'newTargetValue',
-             state)
+            state)
 
     def lock(self):
         """Lock the door."""
@@ -752,6 +729,7 @@ class VeraLock(VeraDevice):
             self.refresh_complex_value('Status')
         val = self.get_complex_value('Status')
         return val == '1'
+
 
 class VeraThermostat(VeraDevice):
     """Class to represent a thermostat."""
@@ -782,7 +760,7 @@ class VeraThermostat(VeraDevice):
             self.thermostat_operating_service,
             'SetModeTarget',
             'NewModeTarget',
-             mode)
+            mode)
         self.set_cache_value('mode', mode)
 
     def get_hvac_mode(self, refresh=False):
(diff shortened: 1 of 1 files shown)


angoyd:
I have tried your changes and it works like a charm even with the RFXtrx devices, much more clean solution.

pavoni:
@angoyd thanks for testing so quickly. Apart from RFXtrx switches - what other devices do you have?

I tested with Switch / Dimmer / Armable Sensor.

I'll release when the thermostat guys have had a chance to test.

I'll close #37 .

angoyd:
@pavoni Glad I could help. I have Switches / Dimmers / Armable senors (mainly via integration of Envisa Alarm these will probably move to a direct integration to home asssitant) / Temp sensors (mainly via RFXtrx) / Zwave Smoke sensors.  All of those seems to be working.

rabbadab:
Works with my RFXtrx devices I'll add some more devices back to the Vera tonight and can test with them later. Great work!

dpembo:
I've tested with z-wave motion sensors, and RFXTrx.  All seems to work great from the HA UI, but when I tried putting automations in to control the RFXTrx devices, it all got confused as 'turning off' devices using the serivce "homeassistant.turn_off"
 that were already off, toggled them on in HA, but left them off in Vera, and they then never re-synced and couldn't be toggled in HA without something changing the state in Vera first.

Never tried this before the changes so couldn't say whether this was an existing problem before I tested them, but the issue affected RFXTrx devices and the z-wave sensor (arm/disarm toggle) devices I currently have setup.

I've also got Harmony integration in Vera I can re-enabled in HA to test also, which I don't really use as I used the direct stuff in HA, but can test this.

pavoni:
@dpembo Sounds very odd. 

So when you call `homeassistant.turn_off` on a device that is `off` in both `homeassistant`  and `vera` it should call the code to turn the vera device `off` - but that shouldn't have any impact.

Can you describe exactly what you do - and what the effect is on both Vera and HA? I've seen bugs before where if the HA state gets out of step with the actual device then you can no longer turn the device on or off - so it sounds like that type of problem.

Happy to take a look if you can point me in the right direction.

dpembo:
it was pretty straight forward - I had an automation rule that when run simply calls homeassistant.turnoff as follows:

`- alias: Away Mode Activated
  hide_entity: true
  trigger:
    platform: state
    entity_id: input_select.home_mode
    # Optional
    from: 'Home'
    to: 'Away'

    # Alias for 'to'
    state: 'Away'

    # If given, will trigger when state has been the to state for X time.
    #for:
    #  hours: 0
    #  minutes: 5
    #  seconds: 0
  #condition:
  #  condition: state
  #  entity_id: input_select.home_mode
  #  state: 'Away'
  action:
    - service: homeassistant.turn_off
      entity_id: light.livingroom_light
    - service: homeassistant.turn_off
      entity_id: light.stairs_light
`
(plus more switches/lights)

Basically, when this automation runs - all devices in the list toggle rather than turn off, so if they're off, they toggle on.   This then results in the device in HA being out of step with Vera - i.e. vera is showing the right state, but HA shows the wrong state.

For Example
LivingRoom Light on Stairs Light off (Both correct in Vera & HA)
Active the automation - the living room light turns off.
In VERA UI - both are off.
In HA UI - Living Room Light is off, Stairs Light is on.

If the device is already off in HA and the command issued from the automation, then the switch in the HA interface toggles on, but the off command to vera is issued (so if I turn on the device device manually, it will turn off even though HA shows it as on). 

At that point, they're out of step with the Vera interface, and the only way to resurrect is to either toggle on in vera to match the state of HA toggles, or to restart HA and it'll re-pickup the correct status from vera again.

That make sense?

pavoni:
@dpembo thanks for the detail. Don't really understand why sending an off command to a device will turn it on in HA. 

HA should be sending an off command (or nothingf it optimises)  to pyvera which should be sending an off command to the Vera hub.

I guess it's possible HA is sending a toggle command for some reason.

That isn't explicitly supported in pyvera - so perhaps something is getting messesd up at that point...

dpembo:
On doing a bit more testing, they never toggle in the opposite direction, only from off to on if the switch/light was off in the first place... if I run the automation again with toggles messed, the toggles still stay on, but the off command in vera is still correctly fired.

pavoni:
This will need some investigation. It may be that pyvera is returning an incorrect state to ha after the 'turn_off' call. If you're a developer I can tell you where to look to check.

The state updates are asynchronous - so it will take a bit of digging. Afraid this works with my zwave devices - so can't investigate here.

dpembo:
Hmm - I also get the issue with z-wave devices - at least when arming/disarming a fibaro motion sensor I appear to have the same problem.

I am a developer so if you point me in the right direction I'll try and debug it when I get some time and the wife doesn't want to kill me for turning the lights on/off continuously 😃 
Python isn't my best language but it's not too difficult

pavoni:
I just tried calling `homeassistant.turn_off` on one of my off lamps - and it was fine - so must be a little harder to get the right circumstances.

If you fancy looking the place to start is `_event_device` in `subscribe.py` this gets the vera events and does state updates - and dispatches calls back to `ha`.

If you trace the data comming into this call it will hopefully provide some clues.

Lights should show their on/off state using `Status: 0` or `Status:1`.

Just so I know if I broke something - was this the same for you the the previous pyvera release?

dpembo:
Thanks - will start to dig... also, I'll switch back to the previous release shortly to test - haven't used pyvera in automation previously as I needed to get all the RF devices in to make it valuable - prior to this I was using command switches with UDP echos to the LWRF hub that I want to avoid using as I'm now mixing zwave and lwrf together through vera as it's a much cleaner solution.

pavoni:
Thanks for the help.

dpembo:
Further digging and more testing with my z-wave device.  When I issue the off and it is already off, logbook in HA shows the device being turned on, even though the vera on command is not issued.  I'll keep you posted.

dpembo:
Other thing I've noticed is that I tend to get 5 or so calls a second going to:

```
GET /data_request?dataversion=52672248&minimumdelay=200&id=lu_sdata&loadtime=1484052669 HTTP/1.1
Host: xxx.xxx.x.xxx:3480
Accept-Encoding: gzip, deflate
Connection: keep-alive
User-Agent: python-requests/2.12.4
Accept: */*

```
Don't recall seeing this level of calls previously?

Am switching back to previous version now.

dpembo:
Switching back - I see much less HTTP requests, more like one of those per 30 seconds rather than 5 per second.... but... I still have the same issues with the switches/toggles and automation 😢 

Even with the previous version of pyvera it still has an issue, so I guess this is a more fundamental HA issue rather than a pyvera one, and other than the very large number of requests to the vera hub (about 300 in a minute) pyvera does seem to work ok.

pavoni:
I'm also running 35.3. There shouldn't be anything like that numbers of calls to the Vera box. The subscription update  should use a long poll - which waits between 200ms and 20s.

That code is in get_changed_devices in __init__.py if you want to investigate.

dpembo:
ok, just trying to figure out whats going on....

Arm is off...

Fire the automation
`17-01-10 22:40:36 homeassistant.components.automation: Executing Away Mode Activated`

Message on the bus
```
17-01-10 22:40:36 homeassistant.core: Bus:Handling <Event call_service[L]: service_data=entity_id=['switch.3_in_1_sensor'], domain=switch, service=turn_off, service_call_id=140279827414656-15>
17-01-10 22:40:36 homeassistant.core: Bus:Handling <Event call_service[L]: service_data=entity_id=['switch.3_in_1_sensor'], domain=switch, service=turn_off, service_call_id=140279827414656-16>
17-01-10 22:40:36 pyvera: Result of data_request with payload {'id': 'lu_action', 'action': 'SetArmed', 'serviceId': 'urn:micasaverde-com:serviceId:SecuritySensor1', 'newArmedValue': 0}: { "u:SetArmedResponse": { "OK": "OK" } }
17-01-10 22:40:36 pyvera: Result of data_request with payload {'id': 'lu_action', 'action': 'SetArmed', 'serviceId': 'urn:micasaverde-com:serviceId:SecuritySensor1', 'newArmedValue': 0}: { "u:SetArmedResponse": { "OK": "OK" } }

```
Then I see this...

`17-01-10 22:40:36 homeassistant.core: Bus:Handling <Event state_changed[L]: entity_id=switch.3_in_1_sensor, new_state=<state switch.3_in_1_sensor=on; last_tripped_time=2017-01-10T07:48:13+00:00, friendly_name=3 in 1 sensor, Vera Device Id=68, device_armed=False, battery_level=100%, device_tripped=False @ 2017-01-10T22:40:36.465350+00:00>, old_state=<state switch.3_in_1_sensor=off; last_tripped_time=2017-01-10T07:48:13+00:00, friendly_name=3 in 1 sensor, Vera Device Id=68, device_armed=False, battery_level=100%, device_tripped=False @ `2017-01-10T22:40:21.360094+00:00>>`
`

Which is where the state is then changed from off to on
and finally the call back through the websocket that updates the UI

`17-01-10 22:40:36 homeassistant.components.websocket_api: WS 140278969194592: Sending {'id': 2, 'type': 'event', 'event': {'time_fired': datetime.datetime(2017, 1, 10, 22, 40, 36, 465384, tzinfo=<UTC>), 'data': {'entity_id': 'switch.3_in_1_sensor', 'old_state': <state switch.3_in_1_sensor=off; last_tripped_time=2017-01-10T07:48:13+00:00, friendly_name=3 in 1 sensor, Vera Device Id=68, device_armed=False, battery_level=100%, device_tripped=False @ 2017-01-10T22:40:21.360094+00:00>, 'new_state': <state switch.3_in_1_sensor=on; last_tripped_time=2017-01-10T07:48:13+00:00, friendly_name=3 in 1 sensor, Vera Device Id=68, device_armed=False, battery_level=100%, device_tripped=False @ 2017-01-10T22:40:36.465350+00:00>}, 'origin': 'LOCAL', 'event_type': 'state_changed'}}

`

pavoni:
So the bus event is probably coming from the get_changed_devices call. Sounds like something is going wrong there. Can you trace the data there? Probably some state into is being interpreted incorrectly.

dpembo:
```
just keeps repeating like this:

17-01-10 23:03:19 pyvera: *****************PEMBO
17-01-10 23:03:19 pyvera: payload {'dataversion': 52672584, 'loadtime': 1484052669, 'minimumdelay': 200, 'timeout': 30}
17-01-10 23:03:19 pyvera: payload {'minimumdelay': 200, 'dataversion': 52672584, 'id': 'lu_sdata', 'loadtime': 1484052669, 'timeout': 60}
17-01-10 23:03:19 pyvera: Result {'loadtime': 1484052669, 'comment': '', 'dataversion': 52672584, 'state': -1, 'full': 0, 'Using_2G': 0}
17-01-10 23:03:19 pyvera: Timestamp {'dataversion': 52672584, 'loadtime': 1484052669}
17-01-10 23:03:19 pyvera.subscribe: Poll returned
17-01-10 23:03:19 pyvera.subscribe: No changes in poll interval

17-01-10 23:03:19 pyvera: *****************PEMBO
17-01-10 23:03:19 pyvera: payload {'dataversion': 52672584, 'loadtime': 1484052669, 'minimumdelay': 200, 'timeout': 30}
17-01-10 23:03:19 pyvera: payload {'minimumdelay': 200, 'dataversion': 52672584, 'id': 'lu_sdata', 'loadtime': 1484052669, 'timeout': 60}
17-01-10 23:03:19 pyvera: Result {'loadtime': 1484052669, 'comment': '', 'dataversion': 52672584, 'state': -1, 'full': 0, 'Using_2G': 0}
17-01-10 23:03:19 pyvera: Timestamp {'dataversion': 52672584, 'loadtime': 1484052669}
17-01-10 23:03:19 pyvera.subscribe: Poll returned
17-01-10 23:03:19 pyvera.subscribe: No changes in poll interval

17-01-10 23:03:19 pyvera: *****************PEMBO
17-01-10 23:03:19 pyvera: payload {'dataversion': 52672584, 'loadtime': 1484052669, 'minimumdelay': 200, 'timeout': 30}
17-01-10 23:03:19 pyvera: payload {'minimumdelay': 200, 'dataversion': 52672584, 'id': 'lu_sdata', 'loadtime': 1484052669, 'timeout': 60}
17-01-10 23:03:19 pyvera: Result {'loadtime': 1484052669, 'comment': '', 'dataversion': 52672584, 'state': -1, 'full': 0, 'Using_2G': 0}
17-01-10 23:03:19 pyvera: Timestamp {'dataversion': 52672584, 'loadtime': 1484052669}
17-01-10 23:03:19 pyvera.subscribe: Poll returned
17-01-10 23:03:19 pyvera.subscribe: No changes in poll interval

17-01-10 23:03:19 pyvera: *****************PEMBO
17-01-10 23:03:19 pyvera: payload {'dataversion': 52672584, 'loadtime': 1484052669, 'minimumdelay': 200, 'timeout': 30}
17-01-10 23:03:19 pyvera: payload {'minimumdelay': 200, 'dataversion': 52672584, 'id': 'lu_sdata', 'loadtime': 1484052669, 'timeout': 60}
17-01-10 23:03:20 pyvera: Result {'loadtime': 1484052669, 'comment': '', 'dataversion': 52672584, 'state': -1, 'full': 0, 'Using_2G': 0}
17-01-10 23:03:20 pyvera: Timestamp {'dataversion': 52672584, 'loadtime': 1484052669}
17-01-10 23:03:20 pyvera.subscribe: Poll returned
17-01-10 23:03:20 pyvera.subscribe: No changes in poll interval

17-01-10 23:03:20 pyvera: *****************PEMBO
17-01-10 23:03:20 pyvera: payload {'dataversion': 52672584, 'loadtime': 1484052669, 'minimumdelay': 200, 'timeout': 30}
17-01-10 23:03:20 pyvera: payload {'minimumdelay': 200, 'dataversion': 52672584, 'id': 'lu_sdata', 'loadtime': 1484052669, 'timeout': 60}
17-01-10 23:03:20 pyvera: Result {'loadtime': 1484052669, 'comment': '', 'dataversion': 52672584, 'state': -1, 'full': 0, 'Using_2G': 0}
17-01-10 23:03:20 pyvera: Timestamp {'dataversion': 52672584, 'loadtime': 1484052669}
17-01-10 23:03:20 pyvera.subscribe: Poll returned
17-01-10 23:03:20 pyvera.subscribe: No changes in poll interval

```

dpembo:
only other useful bit I've pulled out so far is from _event
 
```
17-01-10 23:33:50 pyvera.subscribe: ****PEMBO _event device_data {'room': 0, 'id': 68, 'category': 4, 'armedtripped': '0', 'state': -1, 'altid': '2', 'lasttrip': '1484034493', 'comment': '', 'tripped': '0', 'subcategory': 3, 'armed': '0', 'batterylevel': '100', 'parent': 1, 'name': '3 in 1 sensor'}
17-01-10 23:33:50 pyvera.subscribe: ****PEMBO _event device_id 68
17-01-10 23:33:50 pyvera.subscribe: ****PEMBO Device__list is none
```

There's no device_list so _event_device looks like it never gets called - doesn't seem right?

pavoni:
I don't understand why it's polling so fast - there is supposed to be a timeout of 30s if there is no data to return. Which Vera version are you running? Perhaps they changed the timeout parameter in some way.

Your trace a few lines up seemed to show a vera device of 68 was calling back into HA.

But the ones aboves seems not to be - because no subscription has been setup.

State: -1 also looks slightly odd - I think we're expecting 1 or 0. So it's possible that is the source of the odd behaviour - since -1 will probably be interpreted as `true` ie `on`.

dpembo:
I'm running a vera plus with latest UI7 firmware, version1.7.2414

pavoni:
Might be worth a patch to ignore updates where state is -1. If that fixes the light `on` issue we can think about a better way.

I'll double check my polling is still working properly. I upgraded recently.

dpembo:
Guess it could well be the -1 then!  
The polling didn't seem to be an issue with the previous version of pyvera it was more like one every 30 seconds before so its definitely something in the commits... might step back to the previous version again tomorrow and take a look at what that is doing again in my environment.

dpembo:
re the polling -  the delay in vera still works if I tweak the minimumdelay value - it's just that every poll from pyvera has a minimum delay of 200ms which is why I'm getting 5 calls every second.

dpembo:
ok - looks like issue with the polling is the dataversion number always remains the same as far as I can see and isn't being updated on each call, which is why it reponds immediately.  On tests outside of pyvera it still works correctly with the delays if the dataversion changes.  Next is to work out why that isn't happening!

pavoni:
So on the face of it that means that Vera has some data to report every 200ms! 

Looking at `_event` - it's worth tracing the `device_data_list` parameter, which should contain all the news that Vera has from the devices. 

Looking again `state: -1` means 'no job` while `state: 1` is `job_in_progress'`. 

What that code tries to do is ignore the various in progress states, but trigger an update when the Vera task is complete. 

You can see I added a test to handle zwave devices where there is a state of `no job` but a comment that the job is waiting to start.

My guess is that the chain of events is slightly different for some of your devices - and the code is getting this wrong.

Tracing at the `_event` level might be easier to interpret.

dpembo:
interestingly - I can't recreate the polling problem outside of pyvera.  The timeout and minimumdelay always seem to apply.

dpembo:
So - the timeout isn't being set which is why it's polling like mad... watching the traffic, the GET call doesn't have the timeout in.... and doesn't look like its ever in.
GET /data_request?loadtime=1484052669&id=lu_sdata&minimumdelay=200&dataversion=52675803

pavoni:
@dpembo You've found it - there was a recent refactor of how we call the vera api in data_request. It has a bug that isn't passing the timeout value - but instead removing it and using it as the timeout to the request.

dpembo:
phew :+1:

pavoni:
@dpembo I'm not at home with the vera - but I've pushed a code change that I think will fix it. I'll test later.

dpembo:
ok - I'm getting rather than popping the timeout and polling issue now seems to be sorted!
next to test the automation/switches to see if that was causing any unusual impact...

even if it isn't, I can now see whats going on more easily without the noise :+1:

pavoni:
@dpembo Sorry you spent all that time finding the wrong bug. As you say should be much less noisy - but suspect we still have your real issue to find.

Thanks so much for all your time and help.

dpembo:
Other issue is still there (no surprises I guess).... 
I'll start to try and debug and make sense of that next.

pavoni:
@dpembo My suspicion is the pyvera is misunderstanding something that comes back from the async status updates. Now you're not drowning in data - those traces should provide some helpful data. As you can see from the code - there are a few work-rounds (checking for locks, searching in comments!).

pavoni:
The changes I posted this morning seem to fix the polling bug you found. Thanks again.

Fortunately we didn't update HA to use the broken version.

dpembo:
Excellent - getting back to the original issue, I've simplified my whole vera devices list, removed all automation and got down to just one, and tested again.

Results are the same, but the volume of noise being less means it should be easier to dig.

This is what I see with my 3in1 zwave sensor being triggered on in the HA UI incorrectly folowing the automation of a switch off when it's already off...

```
17-01-12 07:11:54 homeassistant.components.automation: Executing Away Mode Activated
17-01-12 07:11:54 homeassistant.core: Bus:Handling <Event logbook_entry[L]: entity_id=automation.away_mode_activated, message=has been triggered, name=Away Mode Activated, domain=automation>
17-01-12 07:11:54 homeassistant.helpers.script: Script Away Mode Activated: Running script
17-01-12 07:11:54 homeassistant.helpers.script: Script Away Mode Activated: Executing step call service
17-01-12 07:11:54 homeassistant.core: Bus:Handling <Event call_service[L]: service_data=entity_id=['switch.3_in_1_sensor'], service=turn_off, service_call_id=140638238392952-2, domain=homeassistant>
17-01-12 07:11:54 homeassistant.core: Bus:Handling <Event service_executed[L]: service_call_id=140638238392952-1>
17-01-12 07:11:54 homeassistant.core: Bus:Handling <Event call_service[L]: service_data=entity_id=['switch.3_in_1_sensor'], service=turn_off, service_call_id=140638238392952-3, domain=switch>
**17-01-12 07:11:54 pyvera: Result of data_request with payload {'action': 'SetArmed', 'id': 'lu_action', 'newArmedValue': 0, 'serviceId': 'urn:micasaverde-com:serviceId:SecuritySensor1'}: { "u:SetArmedResponse": { "OK": "OK" } }**
17-01-12 07:11:54 homeassistant.core: Bus:Handling <Event state_changed[L]: old_state=<state switch.3_in_1_sensor=off; friendly_name=3 in 1 sensor, battery_level=100%, Vera Device Id=68, last_tripped_time=2017-01-11T23:01:57+00:00, device_tripped=False, device_armed=False @ 2017-01-12T07:00:29.130688+00:00>, new_state=<state switch.3_in_1_sensor=on; friendly_name=3 in 1 sensor, battery_level=100%, Vera Device Id=68, last_tripped_time=2017-01-11T23:01:57+00:00, device_tripped=False, device_armed=False @ 2017-01-12T07:11:54.422923+00:00>, entity_id=switch.3_in_1_sensor>
17-01-12 07:11:54 homeassistant.components.websocket_api: WS 140637563616168: Sending {'event': {'time_fired': datetime.datetime(2017, 1, 12, 7, 11, 54, 422959, tzinfo=<UTC>), 'event_type': 'state_changed', 'data': {'old_state': <state switch.3_in_1_sensor=off; friendly_name=3 in 1 sensor, battery_level=100%, Vera Device Id=68, last_tripped_time=2017-01-11T23:01:57+00:00, device_tripped=False, device_armed=False @ 2017-01-12T07:00:29.130688+00:00>, 'new_state': <state switch.3_in_1_sensor=on; friendly_name=3 in 1 sensor, battery_level=100%, Vera Device Id=68, last_tripped_time=2017-01-11T23:01:57+00:00, device_tripped=False, device_armed=False @ 2017-01-12T07:11:54.422923+00:00>, 'entity_id': 'switch.3_in_1_sensor'}, 'origin': 'LOCAL'}, 'id': 2, 'type': 'event'}

```

Looks like it might be down to the response possibly in pyvera maybe:
`**17-01-12 07:11:54 pyvera: Result of data_request with payload {'action': 'SetArmed', 'id': 'lu_action', 'newArmedValue': 0, 'serviceId': 'urn:micasaverde-com:serviceId:SecuritySensor1'}: { "u:SetArmedResponse": { "OK": "OK" } }**`
which is where I'll start to dig, but I have the same issue with the RFTRX devices also as well as this zwave one.  I haven't tried the harmony stuff yet and have deleted that entirely to keep whats in vera simple right now.

pavoni:
That certainly looks like a problem.

Armed devices are done oddly in pyvera/HA - in that two devices are created for one device - both with the same id.

The subscription update code probably doesn't cope well with multiple devices with the same id.

pavoni:
It may be better to change how this works. Create a single pyvera device and two HA objects.

The HA objects can then be a little smarter about how they react.

dpembo:
Just tested on another device a light switch (dimmer).  Same sort of issue

```
17-01-13 06:38:46 pyvera: Result of data_request with payload {'serviceId': 'urn:upnp-org:serviceId:Dimming1', 'newLoadlevelTarget': 0, 'action': 'SetLoadLevelTarget', 'id': 'lu_action'}: { "u:SetLoadLevelTargetResponse": { "OK": "OK" } }
17-01-13 06:38:46 homeassistant.core: Bus:Handling <Event state_changed[L]: entity_id=light.study_light, new_state=<state light.study_light=on; supported_features=1, brightness=0, Vera Device Id=53, friendly_name=Study Light @ 2017-01-13T06:38:46.591560+00:00>, old_state=<state light.study_light=off; supported_features=1, Vera Device Id=53, friendly_name=Study Light @ 2017-01-13T06:38:04.967792+00:00>>
```

dpembo:
Ok - looks like an issue with the sensor I think

When the automation fires on the light I see:

```
17-01-13 07:52:01 pyvera: **** PEMBO get_value <pyvera.VeraDimmer object at 0x7fc77c739710> level****
17-01-13 07:52:01 pyvera: **** PEMBO get_value <pyvera.VeraDimmer object at 0x7fc77c739710> BatteryLevel****
17-01-13 07:52:01 pyvera: **** PEMBO get_value <pyvera.VeraDimmer object at 0x7fc77c739710> Armed****
17-01-13 07:52:01 pyvera: **** PEMBO get_value <pyvera.VeraDimmer object at 0x7fc77c739710> Tripped****
17-01-13 07:52:01 pyvera: **** PEMBO get_value <pyvera.VeraDimmer object at 0x7fc77c739710> Watts****
```

There's some confusion somewhere as the Dimmer is showing get value requests for the 3 in 1 device.
I think this issue already existed before this pull request by the way.

pavoni:
As you say I think the problem is unrelated to this PR.

When you send an action in Vera, there are a large number of asynchronous updates to the device ('waiting to send command, " command sent waiting for reply" perhaps "retrying"), before it settles into a stable state.

The Vera subscription code receives all of these, tries to filter the ones that HA doesn't care about, and pass on the useful ones.

Some of the intermediate states have the pre action state, so I filter these out so that the switches in HA don't bounce.

So my guess is there are two areas where the code could be wrong:-

1. Confusion over the two devices attached to an armed device (the subscription mechanism is based on Vera id, and both these devices agree the same id). So perhaps a status meaning 'armed on' is being stored in a switch.

2. Some interim message that is being used, that shouldn't be - or some final message that is being ignored incorrectly - or a mix of the two.

2 could be caused by a difference between Vera devices or firmware

1 could be an armable device with different status than the ones I tested with, or just that armable devices probably haven't been used that much.

I think the key to finding both is tracing the event updates.

The values that you are tracing are returned to HA from the cached state, so they show clearly that the state is wrong - although the damage is done before those calls.

dpembo:
So - I think I might have an answer.... still not entirely sure yet, or how to fix it, hoping you can give me a pointer... so

If the device is off in vera already - but I issue an off command through an automation anyway (which for RF devices make sense as they're only 1 way - i.e. they don't broadcast their state).  The off command is issued to vera, but as it was already off, there's no event that comes back because vera hasn't done anything with that device.  Which is why the state in HA is out of sync.   

The change is set in the cache, then for some reason, which I'm not quite sure of this then results in the toggle of the switch to on in HA, and because of the above, never corrects itself....

you think this is right?

dpembo:
ok - light is definitely confusion between switch/dimmer
homeassistant.turn_off doesn't work on a vera dimmer, but light.turn_off does...

next a switch test, then back to the 3in1 issues..

pavoni:
What you say makes sense - although I don't understand what's setting the HA state to on.
Calling switch_off in HA calls the device's switch_off method and sets the HA state 'optimistically' so even if there isn't a call back the device state would be fine.

It does schedule an ha state update - so if the pyvera cached state is incorrectly set that could be an issue.


Just seen your update - I'll check the optimistic state set in the dimmer...

dpembo:
re the switch

switch.turn_off works in an automation
but not homeassistant.turn_off

pavoni:
Interesting.  I think I'm specific in my automations.

I would imagine it should work. Could be a bug in the ha Vera code (which is mostly mine anyway) - probably worth a question in the HA gitter forum.

dpembo:
@pavoni - thanks for everything - have a much deeper understanding of all this now - really appreciate your time and help with this 😄.

dpembo:
I think i'm wrong now... have recreated it even with the specific on/offs... 😕

dpembo:
ok - back to thinking this is an issue with the async part of pyvera.
All works ok on one device, but if I use an automation to drive a few, then I think confusion sets in with the responses/etc.

dpembo:
reverting back to command line REST calls to VERA for now - will setup another VM to allow me to debug the whole process more easily (including HA, not just pyvera) - without annoying the wife :)

This pull request definitely improves things as the RF switches work now - prior to this they didn't, but pretty sure there is an issue with the state still.

pavoni:
I'll merge this - it fixes a bug we introduced in the previous version - and adds RF support.

Before I release it - could @jnewland @robjohnson189 or @rhooper check it supports their thermostats correctly. 

@dpembo thanks for all the help - raise an issue when you want to do some more digging together.

jnewland:
> Before I release it - could @jnewland @robjohnson189 or @rhooper check it supports their thermostats correctly.

This is working great for me. 👍

pavoni:
Thanks @jnewland - will release and update HA later in the week.

robjohnson189:
Working for me as well

pavoni:
Thanks guys - now released as `0.2.23 `

PR #40, opened 2017-02-08
Handle duplicate names

arjenfvellinga:
I have multiple ceiling lights in different rooms and therefor they are named the same (=Plafondlamp).

HASS maps the Vera devices to entities, but with each restart the mapping changes. So one time it is mapped like this:

```
light.plafondlamp_5	 - Vera Device Id: 66
friendly_name: Plafondlamp
```

and on the next restart it could be:

```
light.plafondlamp_5	 - Vera Device Id: 29
friendly_name: Plafondlamp
```

So with duplicate names grouping for instance is impossible. Adding the Vera `device_id` to the name, duplicate names are prevented.

Although I could manually change the device names in the Vera controller it is not obvious for new HASS users why entities changes with a restart.

I thought of sorting the devices in `get_devices` so the order is fixed and the mapping will always be the same. But that isn't true when a new device is added with a similar name.

I'm not sure this is the right way to go, but it solves my problem. But I thinks it will mess with the current users.

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -228,7 +228,7 @@ class VeraDevice(object):
         if self.json_state.get('deviceInfo'):
             self.category = (
                 self.json_state.get('deviceInfo').get('categoryName'))
-            self.name = self.json_state.get('deviceInfo').get('name')
+            self.name = self.json_state.get('deviceInfo').get('name') + ' ' + str(self.device_id)
         else:
             self.category = ''
 


pavoni:
I can see what you describe would be a problem - but I'm not keen to force the device name change on existing users, most of whom probably won't have devices with the same names.

I think a better plan is to add a device suffix if the names clash (which I think is what hass does). 

I don't know if the order of fetching devices is fixed in vera - but I suspect it is, perhaps even in id order.

I think if you add a new device with the same name as an existing device - then it's reasonable for the naming to be unpredictable - although I suspect this will usually work if we fetch devices in vera_id order (I think new devices usually have greater ids than existing ones).

arjenfvellinga:
I understand you are not keen on a name change. I wouldn't be either ;-) But do you know how many hass users use the vera component? Are there such statistics? You could argue that while they are still on a pre-release version such changes are acceptable.

I also thought about changing the naming within HASS,. Without knowing exact where hass appends the counter on name clashes, I suspect it's part of the core where you don't want component specific exceptions.

Or do you think it's possible to find a solution in [vera.py](https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/vera.py)? 

> although I suspect this will usually work if we fetch devices in vera_id order (I think new devices usually have greater ids than existing one

I think it's not safe to assume this. Maybe it's true for newly added devices, but when you administer your current setup you can get my situation.

At least it could be mentioned in the README.md that users shouldn't use the same name twice for now.

arjenfvellinga:
I gave it some thought, and see now that the problem lies in Hass and not in your component

pavoni:
I maintain both ends of this - so happy for us to resolve either in the hass wrapper - or in pyvera, just don't want to mess things up for existing users who don't have this problem.

The logic could be to look for clashing names - and add device id to any names that aren't unique.

Thinking about it - it's probably best to do this in the hass wrapper as you say. I exclude a lot of devices when I load into hass - suspect this is a common pattern - so would be good to avoid renaming unused devices.

arjenfvellinga:
Yes that sounds like the way, add a duplicate naming handler in vera.py after excluding devices.

Maybe you are more experienced with developing on Hass, but I would like to do this. Ok with you?

And the PR will come back to you, so we could discuss then again ;-)

pavoni:
Of course. Let me know when your HA PR is ready to go and I'll take a look.

I'll close this issue - since we're going to fix in HA.

PR #48, opened 2017-04-11
Scene Controller support

alanfischer:
Heres a pull request that supports Scene Controllers.  The more frequent polling to get it to be responsive isn't ideal but I haven't found an easier way, and probably won't have the time to look into it further, but here are the changes if you are interested in merging them.

Thanks!

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -200,7 +200,7 @@ class VeraController(object):
 
         self.device_services_map = service_map
 
-    def get_changed_devices(self, timestamp, timeout=SUBSCRIPTION_WAIT):
+    def get_changed_devices(self, timestamp):
         """Get data since last timestamp.
 
         This is done via a blocking call, pass NONE for initial state.
@@ -209,7 +209,7 @@ class VeraController(object):
             payload = {}
         else:
             payload = {
-                'timeout': timeout,
+                'timeout': SUBSCRIPTION_WAIT,
                 'minimumdelay': SUBSCRIPTION_MIN_WAIT
             }
             payload.update(timestamp)
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -144,6 +144,8 @@ class VeraController(object):
                     self.devices.append(VeraLock(item, self))
                 elif device_category == 'Thermostat':
                     self.devices.append(VeraThermostat(item, self))
+                elif device_category == 'Scene Controller':
+                    self.devices.append(VeraSceneController(item, self))
             else:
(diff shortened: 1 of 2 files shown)


pavoni:
Thanks @alanfischer. Am I right in thinking that you want to poll the `VeraSceneController` every 5 seconds from HA? If so there is an easier way to do this that doesn't require you to change the subscription code.

HA is set up to poll devices on a regular basis - which vera devices turn off because they update via subscription. However I think the only notification for scenes would be for the `active` property - so subscription basically won't work for the attributes you want.

So if in your home assistant device you override`should_poll` and return `true` - then HA will poll your device without needing any of the subscription changes (you can see the base vera `should_poll` here https://github.com/home-assistant/home-assistant/blob/dev/homeassistant/components/vera.py#L154). Your update method will then get polled.

The default polling interval in HA is set by the base class of each device (10s for a switch) - but it can be overridden (I think the insteon switch does this).

So I think you could simplify the changes - and still get the result you want.

alanfischer:
Thanks for the feedback.  It'll probably be another week before I have a chance, but I'll get the changes together and update the pull request.

pyvera/__init__.py
```diff
@@ -198,7 +200,7 @@ def map_services(self):
 
         self.device_services_map = service_map
 
-    def get_changed_devices(self, timestamp):
+    def get_changed_devices(self, timestamp, timeout=SUBSCRIPTION_WAIT):
```
pavoni: Assuming you no longer need this - could you remove?

pyvera/__init__.py
```diff
@@ -207,7 +209,7 @@ def get_changed_devices(self, timestamp):
             payload = {}
         else:
             payload = {
-                'timeout': SUBSCRIPTION_WAIT,
+                'timeout': timeout,
```
pavoni: Same as above.

pyvera/__init__.py
```diff
@@ -490,6 +495,10 @@ def vera_device_id(self):
         """The ID Vera uses to refer to the device."""
         return self.device_id
 
+    @property
+    def should_poll(self):
```
pavoni: Nice approach - assume you'll use the in ths HA wrapper code?

I think there are a few other components that depend on polling (I think locks, but we should check so these are set correctly in the Vera devices). Basically anything that uses `get_complex_value` will need polling.

alanfischer: I can certainly add the should_poll changes to the locks if you'd like that in this pull.

pavoni:
Thanks @alanfischer. Now released as `0.2.27`

PR #57, opened 2017-06-02
Colored lights

alanfischer:
This adds support for setting rgb values on lights.

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -372,6 +372,17 @@ class VeraDevice(object):  # pylint: disable=R0904
             return
         dev_info[name.lower()] = str(value)
 
+    def set_cache_complex_value(self, name, value):
+        """Set a variable in the local complex state dictionary.
+
+        This does not change the physical device. Useful if you want the
+        device state to refect a new value which has not yet updated drom
+        Vera.
+        """
+        for item in self.json_state.get('states'):
+            if item.get('variable') == name:
+                item['value'] = str(value)
+
     def get_complex_value(self, name):
         """Get a value from the service dictionaries.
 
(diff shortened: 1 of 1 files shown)


pyvera/__init__.py
```diff
@@ -611,6 +626,50 @@ def set_brightness(self, brightness):
             percent)
         self.set_cache_value('level', percent)
 
+    def get_color_index(self, colors, refresh=False):
+        """Get color index.
+
+        Refresh data from Vera if refresh is True, otherwise use local cache.
+        Refresh is only needed if you're not using subscriptions.
```
pavoni: The comment `Refresh is only needed if you're not using subscriptions.` isn't quite true in this case - because subscriptions don't update the complex state cache - so would you mind updating the comment?

pyvera/__init__.py
```diff
@@ -611,6 +626,50 @@ def set_brightness(self, brightness):
...
+            self.refresh_complex_value('SupportedColors')
+        sup = self.get_complex_value('SupportedColors').split(',')
+        rgb = [sup.index(c) for c in colors]
+        return rgb
+
+    def get_color(self, refresh=False):
+        """Get color.
+
+        Refresh data from Vera if refresh is True, otherwise use local cache.
+        Refresh is only needed if you're not using subscriptions.
```
pavoni: Same applies here with the comment.

alanfischer:
I've corrected the comments.  I also double checked that the light color information was not in deviceInfo, and sadly it isn't.  Only the light level is there.
I do wish I had lights with other coloursets to test with, if I get my hands on some I can certainly make the changes to support those too.

pavoni:
Thanks @alanfischer. I see you're on a roll (https://github.com/pavoni/pyvera/pull/59).

Will do a release when we've merged that one too.

PR #58, opened 2017-06-03
Fixed fan mode

alanfischer:
According to here: http://wiki.micasaverde.com/index.php/Luup_Devices and my thermostat, this fixes the fan control :)


pavoni:
Thanks.

PR #59, opened 2017-06-07
Temperatures now come back as floats instead of strings

alanfischer:
Home Assistant gets confused when temperatures are returned as strings, this fixes it.


pavoni:
Looks good.

pavoni:
Now released as 0.2.32. Thanks for your contributions.

PR #60, opened 2017-06-11
Return None instead of except when color is not supported

alanfischer:
Now that i got my hands on a non-color dimmer, I see my rgb methods raise exceptions when used on it. 

If you want this patch, this cleans it up so they just return None instead of an AttributeError when trying to split a None.


pavoni:
Thanks - now released as 0.2.33.

Issue #62, opened 2017-06-22
Exception relating to RGB


Whytey:
Thrown from Home-Assistant but appears to be in the pyvera library.
```2017-06-22 21:16:07 ERROR (MainThread) [homeassistant.components.light] Error while setting up platform vera
Traceback (most recent call last):
  File "/usr/local/lib/python3.4/dist-packages/homeassistant/helpers/entity_component.py", line 161, in _async_setup_platform
    SLOW_SETUP_MAX_WAIT, loop=self.hass.loop)
  File "/usr/lib/python3.4/asyncio/tasks.py", line 377, in wait_for
    return fut.result()
  File "/usr/lib/python3.4/asyncio/futures.py", line 275, in result
    raise self._exception
  File "/usr/lib/python3.4/concurrent/futures/thread.py", line 54, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.4/dist-packages/homeassistant/components/light/vera.py", line 24, in setup_platform
    VeraLight(device, VERA_CONTROLLER) for device in VERA_DEVICES['light'])
  File "/usr/local/lib/python3.4/dist-packages/homeassistant/helpers/entity_component.py", line 331, in schedule_add_entities
    self.async_schedule_add_entities, list(new_entities),
  File "/usr/local/lib/python3.4/dist-packages/homeassistant/components/light/vera.py", line 24, in <genexpr>
    VeraLight(device, VERA_CONTROLLER) for device in VERA_DEVICES['light'])
  File "/usr/local/lib/python3.4/dist-packages/homeassistant/components/light/vera.py", line 35, in __init__
    VeraDevice.__init__(self, vera_device, controller)
  File "/usr/local/lib/python3.4/dist-packages/homeassistant/components/vera.py", line 144, in __init__
    self.update()
  File "/usr/local/lib/python3.4/dist-packages/homeassistant/components/light/vera.py", line 86, in update
    self._color = self.vera_device.get_color()
  File "/var/opt/homeassistant/deps/pyvera/__init__.py", line 652, in get_color
    ci = self.get_color_index(['R', 'G', 'B'], refresh)
  File "/var/opt/homeassistant/deps/pyvera/__init__.py", line 640, in get_color_index
    ci = [sup.split(',').index(c) for c in colors]
  File "/var/opt/homeassistant/deps/pyvera/__init__.py", line 640, in <listcomp>
    ci = [sup.split(',').index(c) for c in colors]
[...]

PR #63, opened 2017-06-23
Catch errors when trying to parse RGB colors

alanfischer:
This should catch the R,G,B parsing errors that are being seen, and get_color should just return None, causing Home Assistant to treat it just like a dimmer, not a colored light.

Should fix https://github.com/pavoni/pyvera/issues/62

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -634,14 +634,15 @@ class VeraDimmer(VeraSwitch):
         if refresh:
             self.refresh_complex_value('SupportedColors')
 
-        ci = None
         sup = self.get_complex_value('SupportedColors')
-        if sup is not None:
-            try:
-                ci = [sup.split(',').index(c) for c in colors]
-            except (TypeError, ValueError, IndexError):
-                pass
-        return ci
+        if sup is None:
+            return None
+
+        sup = sup.split(',')
+        if not set(colors).issubset(sup):
+            return None
+
+        return [sup.index(c) for c in colors]
 
     def get_color(self, refresh=False):
         """Get color.
@@ -651,16 +652,13 @@ class VeraDimmer(VeraSwitch):
         if refresh:
             self.refresh_complex_value('CurrentColor')
 
-        rgb = None
         ci = self.get_color_index(['R', 'G', 'B'], refresh)
         cur = self.get_complex_value('CurrentColor')
-        if ci is not None and cur is not None:
-            try:
-                val = [cur.split(',')[c] for c in ci]
-                rgb = [int(v.split('=')[1]) for v in val]
-            except (TypeError, ValueError, IndexError):
-                pass
-        return rgb
+        if ci is None or cur is None:
+            return None
+
+        val = [cur.split(',')[c] for c in ci]
+        return [int(v.split('=')[1]) for v in val]
 
     def set_color(self, rgb):
         """Set dimmer color.
@@ -674,6 +672,9 @@ class VeraDimmer(VeraSwitch):
             target)
 
         rgbi = self.get_color_index(['R', 'G', 'B'])
+        if rgbi is None:
+            return
+
         target = ('0=0,1=0,' +
                   str(rgbi[0]) + '=' + str(rgb[0]) + ',' +
                   str(rgbi[1]) + '=' + str(rgb[1]) + ',' +
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -637,7 +637,10 @@ class VeraDimmer(VeraSwitch):
         ci = None
         sup = self.get_complex_value('SupportedColors')
         if sup is not None:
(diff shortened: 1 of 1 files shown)


pyvera/__init__.py
```diff
@@ -637,7 +637,10 @@ def get_color_index(self, colors, refresh=False):
         ci = None
         sup = self.get_complex_value('SupportedColors')
         if sup is not None:
-            ci = [sup.split(',').index(c) for c in colors]
+            try:
```
Whytey: The list comprehension was making my head hurt, but the fix is quite simple I think:
`ci = [sup.split(',').index(c) for c in colors if c in a]`

tested in the console below:
```
>>> sup
'W,A,I'
>>> colors
['R', 'G', 'B']
>>> ci = [sup.split(',').index(c) for c in colors if c in a]
>>> print ci
[]
```

This doesn't require any exception handling to be added anywhere else.

Whytey:
This will fix https://github.com/pavoni/pyvera/issues/62

pyvera/__init__.py
alanfischer: This works, however the result doesn't tell you which index matched with incoming supported color list.
Example:
`
sup = 'A,I,G'
colors = ['R','G','B']
ci = [sup.split(',').index(c) for c in colors if c in sup.split(',')]
print(ci)
[2]
`
You know something matched at 2, but which one it was is unknown.
My vote would be to just catch any errors if something is not found in the SupportedColors list.  I suppose a 'proper' fix would be to add a get_supported_colors method, so you could first find out of RGB is even supported before trying to get the index of the colors.  Or modifying it so it would insert a `-1` or `None` if an index is unfound.
But if your lights support other color sets then RGB I'm not sure how that would correspond to Home-Assistant.
Is 'W,A,I' what your SupportedColors come back as?

Whytey:
Yes. One device is an Aeon Micro-Dimmer and reports 'W,A' the other is an
Aeon 2-channel Micro-Switch and reports 'W,A,I'.

Whytey:
I tried to find out what the valid values are for SupportedColors and what these other values might indicate but came up empty.

Perhaps then, the fix is to ensure that "R,G,B" is somewhere in SupportedColors first and if not, just return None.

Insert following check before the list comprehension:
```
if not set(colors).issubset(sup.split(',')):
    return None
```

pavoni:
Back from vacation - what's the current thinking on the best way to go here?

alanfischer:
We could do this:

`
def get_color_index(self, colors, refresh=False):
    """Get color index.

    Refresh data from Vera if refresh is True, otherwise use local cache.
    """
    if refresh:
        self.refresh_complex_value('SupportedColors')

    ci = None
    sup = self.get_complex_value('SupportedColors')
    if sup is not None:
        try:
            sup = sup.split(',')
            if set(colors).issubset(sup):
                ci = [sup.index(c) for c in colors]
        except (TypeError, ValueError, IndexError):
            pass
    return ci
`

I still feel we should keep the try/except in there anyway, since I've gotten burned a couple times by the vera hub returning not exactly what I expected it to.  However this would at least check for the colors existing in sup before getting the indexes.

If we go with the above, then I should also make a similar change to get_color since it does similar string splitting.  Thoughts @pavoni or @Whytey?

Whytey:
My main concern was about passing exceptions to method callers, which you are no longer doing.

That said, the try/except appears fairly redundant in this case.  None of those Error types should be thrown, regardless of what Vera returns.

pavoni:
Perhaps slightly nicer to do the checks as guards
```
def get_color_index(self, colors, refresh=False):
    if refresh:
        self.refresh_complex_value('SupportedColors')
    sup = self.get_complex_value('SupportedColors')
    if sup is None:
        return None;
    try:
        sup = sup.split(',')
        if not set(colors).issubset(sup):
            return None
         return [sup.index(c) for c in colors]
    except (TypeError, ValueError, IndexError):
        return None
```
While 'try it and catch the exceptions' is the best way to do things in Python - I have had some bad experiences with code that caught broad exceptions. You can end up catching things that weren't what you expected.

So let's catch things where we know the cause - but not otherwise. Easy to do another patch if we see an exception we didn't expect.

alanfischer:
Does this look good to everyone?  I've removed the try/except since those errors shouldn't be caught, and the checks are now guards.

Whytey:
I don't have any problems with your update @alanfischer .  I am no expert though ;)

jonny190:
seconded @Whytey

pavoni:
Now released as `0.2.34` - and thanks for the contribution and help from everyone.

pavoni:
PS @alanfischer Let me know if you'd like me to do the HA PR - or you'll do that yourself.

alanfischer:
@pavoni You are welcome to do the PR if you don't mind.

alanfischer:
Got the HA PR submitted, so no need to worry about it.

PR #68, opened 2017-07-19
Catch IndexError in get_color

alanfischer:
Catch IndexError exception in get_color


pavoni:
Thanks - released as `0.2.35`.

PR #69, opened 2017-07-23
Fix user config error

### LINKED ISSUE #66: Device with Vera warning does not update
Using this as part of home assistant. 

I have 3 Aeotec 6 in 1 multisensors. For some reason my Vera always displays "failed at setting user config" for these devices. However the settings have updated. 

So the issue is, motion sensor reports stop getting sent to Home Assistant for these devices. The other sensors such as temperature and lux update in home assistant, but the motion detection does not. The motion detection does however report correctly in Vera. 

I have a bunch of other sensors such as the 4 in 1 sensor and door window sensors. All these work perfectly. 

So, not sure why pyvera isn't updating the status when motion is/isn't detected for these motion sensors.

philhawthorne:
If a vera device is unable to set user configurations, it may still
report updated values correctly.

This PR will check if the state is error, and the comment contains the
"Setting user configuration" string, and if it does allow the state to
be updated for the device.

Fixes #66

### DIFF
--- a/pyvera/subscribe.py
+++ b/pyvera/subscribe.py
@@ -86,6 +86,8 @@ class SubscriptionRegistry(object):
         if not (state == STATE_JOB_DONE or
                 state == STATE_NOT_PRESENT or
                 state == STATE_NO_JOB or
+                (state == STATE_JOB_ERROR and 
+                comment.find('Setting user configuration'))):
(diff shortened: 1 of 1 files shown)


pavoni:
Sorry I missed this @philhawthorne - will release a new version soon.

Thanks for the contribution.

PR #71, opened 2017-08-31
Use ids

alanfischer:
Use category ids instead of names


pavoni:
Will merge this now so we can test.

There is still a little `category_name` logic left which we should think through - both for queries (don't know how much this is used) and for Armable sensors.

pavoni:
Should be available as `0.2.36a0` although I tried to call it `0.2.36alpha`

Worth checking if this works for you.

alanfischer:
@pavoni Do you have armable sensors to test with?  I don't own any so I'm not exactly sure what changes should be made to have them behave correctly.

pavoni:
I do have one (although I don't use the arming).

I'll take a look.

PR #74, opened 2017-10-27
Dont create an armable device if its already a switch

alanfischer:
This should resolve this issue:

https://community.home-assistant.io/t/entity-id-already-exists/30305

Where there is a switch thats also armable, so it ends up creating 2 switches in home assistant of the same name.  Looking back at the pre-category-id vera code, it didnt create an armable device for an on/off switch, so this should go back to that behavior.


pavoni:
Thanks for sorting this - now released as 0.2.38.

PR #75, opened 2017-11-04
Added scene support

alanfischer:
This adds scene support

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -989,7 +989,8 @@ class VeraScene(object):
         self.json_state = json_obj
         self.scene_id = self.json_state.get('id')
         self.vera_controller = vera_controller
-        self.name = ''
+        self.name = self.json_state.get('name')
+        self._active = False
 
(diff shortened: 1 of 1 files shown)


alanfischer:
Closing this for now, I realized I need to expose a couple other items for a clean home-assistant integration...

alanfischer:
Ok, this should be better...

pavoni:
Looks good - duplicates some of the `VeraDevice` methods. What was the reason you decided to duplicate rather than inherit?

alanfischer:
The api exposes the scenes separately from the devices:  {"scenes":[{...},...], "devices":[{...},...]}
And the scene ids are a separate set than the device ids, so scene ids overlap with device ids.
If you would prefer I can certainly still use the VeraDevice & device_id members in there to store the scene items.

alanfischer:
Ok, added some comments related to these

pavoni:
Thanks for this - now released as `0.2.39`

PR #79, opened 2017-11-13
Added unicode support for device and scene names.

dahlo:
I noticed that the script broke when running it on my system, which has non-ascii characters in the device names for some devices. I added unicode encoding of the device names and now it runs.

Lovely script btw, this is exactly what Vera should supply themselves instead of the very incomplete wiki page about Luup requests..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ __pycache__
 build
 dist
 pyvera.egg-info
+*.pyc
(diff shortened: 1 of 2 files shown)


dahlo:
Ahh, my bad, i guess you are running Python 3 where this is not a problem. I should get with the times i guess..

This pull request will break the Python 3 support so I would not implement it the way it is now.

PR #80, opened 2017-11-13
Added unicode support for Python 2.x

dahlo:
I noticed that the script broke when running it on my system, which has non-ascii characters in the device names for some devices. I added unicode encoding for the device and scene names when using Python 2.x and now it runs. Running it with Python 3+ works as well.

Lovely script btw, this is exactly what Vera should supply themselves instead of the very incomplete wiki page about Luup requests..

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -9,6 +9,20 @@ import sys
 
 from .subscribe import SubscriptionRegistry
 
+
+# activate to see raw http requests
+debug=0
+if debug:
+    import httplib as http_client
+    http_client.HTTPConnection.debuglevel = 1
+    logging.basicConfig()
+    logging.getLogger().setLevel(logging.DEBUG)
+    requests_log = logging.getLogger("requests.packages.urllib3")
+    requests_log.setLevel(logging.DEBUG)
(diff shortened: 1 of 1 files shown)


pyvera/__init__.py
```diff
@@ -5,9 +5,24 @@
...
+debug=0
+if debug:
+    import httplib as http_client
+    http_client.HTTPConnection.debuglevel = 1
+    logging.basicConfig()
+    logging.getLogger().setLevel(logging.DEBUG)
+    requests_log = logging.getLogger("requests.packages.urllib3")
+    requests_log.setLevel(logging.DEBUG)
+    requests_log.propagate = True
+
```
pavoni: Thanks for the PR - look like a few extra blank lines made it into your debug logic.

I'm slowly trying to improve the code to get through pep8 / pylint - so would you mind checking the new code gets through OK.

Then happy to merge.

dahlo: Oh, forgive me, i'm a bit new to this with pull request. I thought it was tied to the latest commit when issuing the request, but i see now that it takes the master branch as it updates. I'll clean out the code i wrote after the request and make it nice and tidy.

PR #81, opened 2017-11-14
Returns unicode for device and scene name in Python 2.x as well.

dahlo:
There we go, a new pull request on a separate branch containing only the unicode code!


pavoni:
Sorry I missed this - now merged and released as 0.2.40.

Thanks for the contribution.

PR #85, opened 2018-02-12
Added support for built in siren in VeraSecure.

patrik3k:
Added CATEGORY_VERA_SIREN for device category 24 which is the built in siren in the "new" VeraSecure controller. The siren is now implemented as a switch and is not as a generic device. Turning switch on, now turns siren sound on. Off turns sound off.


pavoni:
Looks good - now released as 0.2.41. Thanks very much for the contribution.

I think you're a HA user - would you like to do the HA pull request or shall I?

PR #87, opened 2018-02-22
Poll hardening

toggledbits:
Harden the polling loop by checking the request Response object in some detail to detect various anomalies that can occur during Vera luup reloads. Since not all JSON implementations stick to de-factos for exception handling, and the requests module returns no error on disconnect with no data, widen the net for catching problems and handle them within a smaller subset of exceptions the polling thread and other callers can manage. Streamline the polling thread code itself just a bit, and add some logging in a few places where it may help to address future issues.

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -9,6 +9,7 @@ import sys
 import json
 
 from .subscribe import SubscriptionRegistry
+from .subscribe import PyveraError
 
 __author__ = 'jamespcole'
 
@@ -247,22 +248,25 @@ class VeraController(object):
         r = self.data_request(payload, TIMEOUT*2)
         r.raise_for_status()
 
-        # Unfortunately, not all JSON implementations return the well-known
-        # json.decoder.JSONDecodeError, so attempt to trap and raise what
-        # the poller expects.
+        # If the Vera disconnects before writing a full response (as lu_sdata 
+        # will do when interrupted by a Luup reload), the requests module will 
+        # happily return 200 with an empty string. So, test for empty response,
+        # so we don't rely on the JSON parser to throw an exception.
+        if r.text == "":
+            raise PyveraError("Empty response from Vera")
+            
+        # Catch a wide swath of what the JSON parser might throw, within
+        # reason. Unfortunately, some parsers don't specifically return 
+        # json.decode.JSONDecodeError, but so far most seem to derive what
+        # they do throw from ValueError, so that's helpful.
         try:
             result = r.json()
-        except json.decoder.JSONDecodeError:
-            raise # pass it up
-        except Exception as ex:
-            raise requests.RequestException("JSON decode error: " + str(ex))
+        except ValueError as ex:
+            raise PyveraError("JSON decode error: " + str(ex))
 
         if not ( type(result) is dict 
                  and 'loadtime' in result and 'dataversion' in result ):
-            LOG.debug("Unexpected or empty response from Vera: %s",
-                      str(result))
-            # Return empty response with unmodified timestamp for retry.
-            return [ None, timestamp ]
+            raise PyveraError("Unexpected/garbled response from Vera")
             
         # At this point, all good. Update timestamp and return change data.
         device_data = result.get('devices')
--- a/pyvera/subscribe.py
+++ b/pyvera/subscribe.py
@@ -24,7 +24,11 @@ STATE_NOT_PRESENT = 999
 
 LOG = logging.getLogger(__name__)
 
-
+class PyveraError(Exception):
+    def __init__(self, arg):
+        self.strerror = arg
+        self.args = {arg}
+    
 class SubscriptionRegistry(object):
     """Class for subscribing to wemo events."""
 
@@ -95,7 +99,15 @@ class SubscriptionRegistry(object):
             return
         device.update(device_data)
         for callback in self._callbacks.get(device, ()):
-            callback(device)
+            try:
+                callback(device)
+            except:
+                # (Very) broad check to not let loosely-implemented callbacks
+                # kill our polling thread. They should be catching their own
+                # errors, so if it gets back to us, just log it and move on.
+                LOG.exception(
+                    "Unhandled exception in callback for device #%s (%s)",
+                    str(device.device_id), device.name)
 
     def join(self):
         """Don't allow the main thread to terminate until we have."""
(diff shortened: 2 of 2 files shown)


pyvera/__init__.py
```diff
@@ -241,9 +242,30 @@ def get_changed_devices(self, timestamp):
...
+        r.raise_for_status()
+
+        # Unfortunately, not all JSON implementations return the well-known
+        # json.decoder.JSONDecodeError, so attempt to trap and raise what
+        # the poller expects.
+        try:
+            result = r.json()
+        except json.decoder.JSONDecodeError:
+            raise # pass it up
+        except Exception as ex:
```
pavoni: It's generally considered bad practice to catch all exceptions like this. Mainly because some of what you might catch isn't what you intend (eg this would even catch Python syntax errors!).

This has a very restricted scope - so isn't a disaster - but did you try to be specific and have problems?

pyvera/__init__.py
```diff
@@ -241,9 +242,30 @@ def get_changed_devices(self, timestamp):
...
+            raise requests.RequestException("JSON decode error: " + str(ex))
```
pavoni: This is really a exception of our own devising - rather than a `RequestException` - so I think it would be better to define an exception of our own - rather than repurposing an existing exception.

They are easy to define
```
class VeraDecodeException(Exception):
    pass
```
Although you'll need to import it into `subscript.py` so you can catch it.

Happy to help if needed.

pyvera/subscribe.py
```diff
@@ -123,23 +123,33 @@ def _run_poll_server(self):
...
-            except requests.RequestException:
-                LOG.info("Could not contact Vera - will retry in %ss",
-                         SUBSCRIPTION_RETRY)
-                time.sleep(SUBSCRIPTION_RETRY)
-            except json.decoder.JSONDecodeError:
-                LOG.info("Response was malfomed - will retry in %ss",
-                         SUBSCRIPTION_RETRY)
-                time.sleep(SUBSCRIPTION_RETRY)
+
+                continue
```
pavoni: I think with you're restructuring this is now not doing anything - so can be removed.

pyvera/__init__.py
toggledbits: The specific problem I've run into is that the JSON module that comes bundled with Hass.io does not throw the (somewhat) standard json.decode.JSONDecodeError on parse errors. But I think I've got the preliminary checks prior to parsing tight enough that parse errors would truly be... ahem... exceptional. I left it in, being conservative on pyvera's stability side. Now, not every best or bad practice is best or bad in every circumstance, but I agree it's overly broad now that other work mitigates much of the risk on the fetches. I'll pull it out.

That does, however, introduce a risk from a different direction. After making a successful fetch of data, updates are sent to the "observers" through callbacks, and anything, including any exception, can be raised in those callbacks outside of pyvera's control, but would take pyvera's thread down as a whole. As currently structured, the callbacks are invoked within the same try block as the data fetches, so once the current broad exception is removed, those exceptions arising in a callback will stop the loop. So, I think some additional protection around the invocation of *each* callback is warranted, again for the longevity of the loop, but the question then is, what are you comfortable with? Since a callback can do anything, this is one of those places where I would argue that the widest possible exception trap is perhaps warranted. Again, this is a very limited scope.

What are your thoughts?

pyvera/subscribe.py
toggledbits: Check again. If you remove the continue, the flow will leave the try block and go into the "Could not contact Vera" path of the code that follows the try block--I don't think we want that.

pyvera/__init__.py
toggledbits: Actually, I had this general exception in two places... in the get_changed, and in polling loop itself. My comments above apply to the polling loop. Sorry for the confusion.

pyvera/subscribe.py
pavoni: 👍 You're right, sorry.

pyvera/subscribe.py
```diff
@@ -123,23 +123,33 @@ def _run_poll_server(self):
...
+
```
pavoni: Just noticed you deleted my `time.sleep(1)`. 

It was quite a while back - but I think I found a case where it was possible to kill vera by polling too much when there were a lot of events - so I ended up using short timeout for the poss (so in most cases HA responds quickly) - along with this sleep to slow things down during very busy times.

pyvera/__init__.py
pavoni: I don't really have strong views - except that I think it's important to trace unexpected errors. 

Had a terrible time debugging some code where someone had caught everything - and something deep in the code sometimes bumped into an undefined variable. The code caught this silently!

So fine to silently catch the things you know about - but lets trace the unexpected.

pyvera/subscribe.py
```diff
@@ -123,23 +123,33 @@ def _run_poll_server(self):
...
+                LOG.debug("Caught JSONDecodeError: %s", str(ex))
+                pass
+
             except Exception as ex:
-                LOG.exception("Vera thread exception %s", ex)
-                raise
+                LOG.exception("Vera poll thread general exception: %s", 
+                              str(ex))
+                # Don't raise. Stay in control
+                #raise
```
pavoni: Let's make a decision - if we're going to catch everything - delete the commented out raise.

My intent with the trace was to add additional exceptions as they came up. But I can see your point above.

pyvera/subscribe.py
toggledbits: Yeah, no problem. HA is very responsive. Moreso than Vera's native UI, so there's obviously goodness in your choices there.

pyvera/subscribe.py
toggledbits: OK. Here's my vote. I think in doing the JSON decode, that ONE statement in its own try block, we should trap anything that comes at up, and report it back as a parsing error in some form, either abusing an existing exception as I did, or introducing a new one. In the loop itself, we can remove the blanket exception, or more to my preference, keep it, message it, but uncomment the raise and let it continue to propagate. If we want to log more in those areas, I'm never opposed to that. I'm also in the "more information is better than less" camp.

WRT the callbacks, I think we should do the same. A callback should be trapping and reporting its own exceptions. If it gets to us, it's serious heartburn, but we should not let that take out the entire thread. On my Vera, there are 150 devices, and one device in particular occasionally gives an HA callback heartburn and knocks the whole thing out: it returns a string longer than 255 characters, and the HA callback chokes somewhere on it. That callback is making two bad choices, one of which is not trapping its own errors. I think we should blanket-catch that ONE statement (the invocation of the callback), log at error level the untrapped (now pyvera-trapped) exception, and continue on.

This gives those two cases VERY limited scope, with a reasonable expectation of good behavior on the part of the parser, and good self-management on the part of the callback. 

If this was my code, I would actually refactor the entire polling loop so that the get_changed call was in its own try small block (to catch whatever is passed on by that call that should not kill the thread--just enough to distinguish temporary errors from likely permanent ones), and the callback invocation was in its own one-statement try block, but everything else was naked, because really, there's nothing else that should be throwing elsewhere in that code, other than latent logic/coding errors or odd exceptions for which you would want a good, untrapped exception to be thrown (with its very accurate traceback associated).

That's my 2p. I'm open to whatever fits with your sensibilities. I am, of course, just trying to make the code harder to tip over.

toggledbits: OK. I dug in a little more, trying to narrow the scope of the polling loop checks as much as possible. From what I've been able to discover, even the loosest of the python JSON parsers likely to be used derives their exceptions from ValueError, so restricting the scope of the parsing catches to that class seems like an acceptable, desirable middleground. I'm testing that now, beating the heck out of my Vera every way I can think of. So far, so good.

toggledbits:
OK. I've now submitted a revision which exception handling on the Vera requests that seem to be quite stable, and without a broad-brush exception trap. I think this is a satisfactory middleground, because as I mentioned in my other comment, although various json parsers are giving mixed responses to errors (not always json.decoder.JSONDecodeError), they seem to all consistently follow the suggestion that they be derived from ValueError. 

In addition, this version adds our own exception class for the above errors, and restructures the polling loop so that its exception trap only contains the code where those exceptions would be expected to be raised. There is also a blanket exception trap bracketing only the callback call, so trap any problem untrapped by the callback itself and log it, but not let it take down the polling thread.

pavoni:
That looks great - now released `0.2.42`.

Thanks for all your hard work on this - I think your solution to this is really well thought through.

Would you like to do the HA PR - or shall I?

toggledbits:
Since HA has a process you've been through, if you don't mind, I'll leave it to you. Thanks for your support and discussions/recommendations on this.

pavoni:
Looks like we made it into 65.3.

Thanks again.

PR #88, opened 2018-04-09
Implement additional attributes for lock types and add a couple extra global vera attributes

jwater7:
Add examples directory and example scripts that list devices and show lock device information.  Implement a "get_strict_value" that does not ignore uppercase characters.  Implement room and comm_failure vera methods.  Add more lock attributes such as last user, pin failure, unauth user, lock failure, and pin code list.

### DIFF
--- a/examples/list-devices.py
+++ b/examples/list-devices.py
@@ -1,0 +1,31 @@
+#!/usr/bin/env python
+
+# Import project path
+import sys
+import os
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.realpath(__file__)), '..'))
+
+# Import pyvera
+import pyvera
+
+# Parse Arguments
+import argparse
+parser = argparse.ArgumentParser(description='list-devices')
+parser.add_argument('-u', '--url', help="Vera URL, e.g. http://192.168.1.161:3480", required=True)
+args = parser.parse_args()
+
+# Start the controller
+controller, _ = pyvera.init_controller(args.url)
+
+try:
+    # Get a list of all the devices on the vera controller
+    all_devices = controller.get_devices()
+
+    # Print the devices out
+    for device in all_devices:
+        print('{} {} ({})'.format(type(device).__name__, device.name, device.device_id))
+
+finally:
+    # Stop the subscription listening thread so we can quit
+    controller.stop()
+
--- a/examples/show-lock-info.py
+++ b/examples/show-lock-info.py
@@ -1,0 +1,40 @@
+#!/usr/bin/env python
+
+# Import project path
+import sys
+import os
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.realpath(__file__)), '..'))
+
+# Import pyvera
+import pyvera
+
+# Parse Arguments
+import argparse
+parser = argparse.ArgumentParser(description='show-lock-info')
+parser.add_argument('-u', '--url', help="Vera URL, e.g. http://192.168.1.161:3480", required=True)
+args = parser.parse_args()
+
+# Start the controller
+controller, _ = pyvera.init_controller(args.url)
+
+try:
+    # Get a list of all the devices on the vera controller
+    all_devices = controller.get_devices()
+
+    # Look over the list and find the lock devices
+    for device in all_devices:
+        if isinstance(device, pyvera.VeraLock):
+            print('{} ({})'.format(type(device).__name__, device.name, device.device_id))
+            print('    comm_failure: {}'.format(device.comm_failure))
+            print('    room_id: {}'.format(device.room_id))
+            print('    is_locked(): {}'.format(device.is_locked()))
+            print('    get_pin_failed(): {}'.format(device.get_pin_failed()))
+            print('    get_unauth_user(): {}'.format(device.get_unauth_user()))
+            print('    get_lock_failed(): {}'.format(device.get_lock_failed()))
+            print('    get_last_user(): {}'.format(device.get_last_user()))
+            print('    get_pin_codes(): {}'.format(device.get_pin_codes()))
+
+finally:
+    # Stop the subscription listening thread so we can quit
+    controller.stop()
+
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -7,6 +7,7 @@ import logging
 import requests
 import sys
 import json
+import os
 
 from .subscribe import SubscriptionRegistry
 from .subscribe import PyveraError
@@ -37,8 +38,15 @@ CATEGORY_UV_SENSOR = 28
 
 _VERA_CONTROLLER = None
 
-LOG = logging.getLogger(__name__)
-
+# Set up the console logger for debugging
+logger = logging.getLogger(__name__)
+# Set logging level (such as INFO, DEBUG, etc) via an environment variable
+# Defaults to WARNING log level unless PYVERA_LOGLEVEL variable exists
+logger.setLevel(os.environ.get("PYVERA_LOGLEVEL", "WARNING"))
+ch = logging.StreamHandler()
+ch.setFormatter(logging.Formatter('%(levelname)s@{%(name)s:%(lineno)d} - %(message)s'))
+logger.addHandler(ch)
+logger.debug("DEBUG logging is ON")
 
 def init_controller(url):
     """Initialize a controller.
@@ -244,7 +252,7 @@ class VeraController(object):
             'id': 'lu_sdata',
         })
         
-        LOG.debug("get_changed_devices() requesting payload %s", str(payload))
+        logger.debug("get_changed_devices() requesting payload %s", str(payload))
         r = self.data_request(payload, TIMEOUT*2)
         r.raise_for_status()
 
@@ -405,7 +413,7 @@ class VeraDevice(object):  # pylint: disable=R0904
             parameter_name: value
         }
         result = self.vera_request(**payload)
-        LOG.debug("set_service_value: "
+        logger.debug("set_service_value: "
                   "result of vera_request with payload %s: %s",
                   payload, result.text)
 
@@ -416,7 +424,7 @@ class VeraDevice(object):  # pylint: disable=R0904
         """
         result = self.vera_request(id='action', serviceId=service_id,
                                    action=action)
-        LOG.debug("call_service: "
+        logger.debug("call_service: "
                   "result of vera_request with id %s: %s", service_id,
                   result.text)
         return result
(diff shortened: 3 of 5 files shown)


jwater7:
@pavoni  - I'd love to see these features incorporated into pyvera since home-assistant uses it and could be modified to get these additional attributes from the lock device.  Not sure what you're process is to get contributions in so let me know if I should do anything else that can help.  Thanks

examples/list-devices.py
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env python
+
+# Import project path
+import sys
+import os
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.realpath(__file__)), '..'))
```
pavoni: Great to include some examples - but why do you need this?

examples/show-lock-info.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python
+
+# Import project path
+import sys
+import os
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.realpath(__file__)), '..'))
```
pavoni: Same question as above.

pyvera/__init__.py
```diff
@@ -466,8 +466,13 @@ def get_value(self, name):
 
         This data is updated by the subscription service.
         """
+        return self.get_strict_value(name.lower())
+
+    def get_strict_value(self, name):
```
pavoni: Can see why this is needed. I added the `lower` because the keys from `get_complex_values` are camel case - but the ones in deviceInfo are _mostly_ lower case.

Whatever we do at this point is ugly - but I wonder if it's better to keep to a single `get_value` call for now - and have it attempt both the passed key - and a lower case value. So something like 

```
    def get_value(self, name):
        """Get a value from the dev_info area.

        This is the common Vera data and is the best place to get state from
        if it has the data you require.

        This data is updated by the subscription service.
        """
        dev_info = self.json_state.get('deviceInfo')
        ret = dev_info.get(name.lower(), None)
        return ret if ret is not None else dev_info.get(name, None)
```

What do you think?

pyvera/__init__.py
```diff
@@ -878,10 +893,89 @@ def is_locked(self, refresh=False):
         Refresh data from Vera if refresh is True, otherwise use local cache.
         Refresh is only needed if you're not using subscriptions.
         """
+        try:
+            if refresh:
+                self.refresh()
+            return self.get_value("locked") == '1'
+        except (TypeError, ValueError):
```
pavoni: Do you know the circumstances where the fallback is needed? Is it just on startup or some sort of race condition?

examples/list-devices.py
jwater7: There’s probably a better way that I’m just not aware of, but this is an attempt to import the project path in a way that is portable so that you can run the script from anywhere and it will find the module appropriately relative to the example file.

examples/show-lock-info.py
jwater7: See above

pyvera/__init__.py
jwater7: That would be fine as well.  Let me know if you’d like me to make that change.

pyvera/__init__.py
jwater7: If I remember correctly, there was no race condition or circumstance that I encountered where the fallback was used.  I believe I added it there for the sake of completeness since it appeared to actually be the only documented “micasa” way of getting the locked status

jwater7:
Glad to see you’re back from vacation, I can dust off my repo again 😁 
Made some code review comments, let me know if you have any additional questions.

pyvera/__init__.py
toggledbits: Want to throw my 2p in here. I think using get_complex_value("Status") is superior to get_value("locked") here and the former should be preferred, not the fallback. The reason is that in the absence of Refresh=True, they are about equal, but with Refresh=True, the sdata refresh needed with get_value() requires fetching of data for ALL devices (an sdata request always returns full data) and then a linear search through those results to find the subject device, where the complex value refresh uses a more efficient single-device ("status" with DeviceNum parameter) request.

However, and I'm willing to tackle this as a separate PR, it's also possible that the sdata refresh() could be made much more efficient by incorporating the use of get_changed_devices() and applying the results to the currently cached data. Additionally, if the polling/subscription loop is running, it could be updating the pyvera's cached sdata as it works, something it does not currently seem to do. Right now, the polling/subscription loop could easily provide a different value to a client callback than a get_value() call (without prior refresh) would return for the same name, and that should not be the case. Addressing this would make the need for calls with Refresh=True much less necessary/likely, and reduce load on the Vera and efficiency for the client.

pyvera/__init__.py
toggledbits: I think this should be formalized as separate, not homogenized, and here's what I mean.

The service state values, which come from the status request to Vera, are the master set of device values and states. These are the Pascal-case names.

The shortcode values, which come from the sdata request to Vera, are a small subset of the service state that is defined by each device's service file (S_device.XML). 

I think get_complex_value() should be deprecated and a new method added named get_service_state() or get_service_value(), which is more in keeping with Vera's/UPnP's terminology and familiar to the Vera developer community, and get_value() should be deprecated and a new method added called get_shortcode_value() or similar. 

Vera clients should be well aware of the differences between these values. Combining them into a single get_value() call creates the opportunity for collisions, and more confusion than this odd disconnect by Vera already creates. 

It's worth nothing that the status request (full state variable results) can also accept the DataVersion parameter, to return incremental changes to prior requests, as lu_sdata does in get_changed_devices(). In my perfect world, pyvera would not have used the shortcode/sdata here, but rather status, thus making the full suite of changes a device could see available to subscribers. As it stands today, only the subset is published.

jwater7: Some really good points toggledbits, I would also prefer the full suite for status returned, maybe this is worth documenting as a feature request?

If the end goal is to make these non-homogeneous, then maybe it would be worth renaming these functions now as you describe and then we make another pull request to split them as you described?

pyvera/__init__.py
jwater7: I hadn’t considered that, we might see some good improvements with that scheme.  We may make a sacrifice of commonality between devices for better performance right?
If you write up the pr, I’d be willing to help out where I can.  You could start with some pseudo code or an example to help better explain your vision and then we could latch on

pyvera/__init__.py
toggledbits: I would not rename them, as that would break existing clients. When we get around to implementing the new names, just comment the old methods "deprecated" (leave them in place and functional) and give the world a long while to change to the new methods before removing the old.

pyvera/__init__.py
pavoni: Given the fallback isn't needed - I suggest we remove it to match the rest of the code base for now. If we want to move away from the `sdata` calls we can do this as another PR.

For now I think `sdata` is the better approach because it works with the subscription mechanism, and most users of pyvera are using it as part of HA - and so will benefit from realtime updates. I accept that there may be a better way (PR welcome).

This assumes that the subscription mechanism works correctly for locks (you might want to check that the ugly code here (https://github.com/pavoni/pyvera/blob/master/pyvera/subscribe.py#L78) works and is still needed, and subscriptions work properly for locks.

pyvera/__init__.py
jwater7: Ok - left as-is for now, can work it later

jwater7:
Ok @pavoni @toggledbits, I made an update that removed the get_complex_value callbacks and also added some better error handling and logging as well (note: I also changed the logger name to not get confused with constants).  There are a couple more commits in there just because I was fiddling trying to get upstream master changes merged in so there are no merge conflicts.  Let me know if there's anything else you'd like before you'd like to merge.  Thanks

pavoni:
This looks good to me - any comments before I merge @toggledbits?

toggledbits:
As the code is today, I think it would have been preferable to make get_complex_value() the lead request, and get_value() the fallback, for the performance reasons I stated. But, if we do the refresh() optimization I suggested (and I certainly will take that on), that minor issue is temporary. We're good to go, as far as I'm concerned. Thanks for great discussion once again!

pavoni:
@toggledbits At this point with a subscription running `get_value` gets event driven realtime data without polling the device - so I think it's superior to `get_complex_value` which requires polling - and doesn't get event updates.

Most users will be running HA - so they should benefit from the subscription mechanism.

So I think the `get_value` path is better at this point.

toggledbits:
@pavoni If the device is under subscription, it should not be necessary to call get_value() at all, because the value is being passed to the subscriber's callback when the device is updated. But the subscription loop does not update the cached copy of the sdata that the VeraDevice object is holding--it doesn't share its newly-gotten update with the cache--so the two will be out of sync at some point, and a client that separately calls get_value() is likely to get a different answer than a subscription callback for the same device passed shortly before. So, in effect, get_value() also requires polling (via Refresh=True) to keep consistent data with a subscription Neither get_value() nor get_complex_value() can be trusted, as Refresh=True must be used to bring them in sync. And so I view them as equals currently, except that the refresh function for get_value() takes a lot more processing (fetches all device data, and then has to search through it) than get_complex_value() (queries for a single device, shorter response and no search required).

But as I said, my intention is to fix the caching/change fetch so that it stays in sync with what a subscriber would see, and hopefully mostly obviates the need to call either get_value() or get_complex_value() (or its eventual successors if renamed) with Refresh=True. It is also the case that both sets of names (the real state names and the short codes) can be kept in sync in this way. It will take a little more work, but it is readily possible with available responses from Vera.

pavoni:
I thought that the `deviceInfo` parts of the `json_state` were updated by the subscription.

See
https://github.com/pavoni/pyvera/blob/master/pyvera/subscribe.py#L98
and
https://github.com/pavoni/pyvera/blob/master/pyvera/__init__.py#L508

Slightly hard to remember which languages return mutable dictionaries - but I think Python does - so I think this will mutate the local state. I'm pretty sure I would have checked this when I wrote the subscription code.

Have you had a case where it wasn't updating?

jwater7:
@pavoni @toggledbits anything stopping us from merging this pull request and then continuing the conversation at #91?  It would be super sweet to get this and the other merge request in and then we can keep building on it.

pavoni:
@jwater7 Happy to merge - I should test against my system before we release - but I think we'll merge some more first.

jwater7:
@pavoni great, thanks!  Yes, that makes sense to wait for a few more changes before release.  Can you take a look at the other pull request and let me know what you think?

toggledbits:
@pavoni nothing stopping the proposed changes right now. Sorry, I said that earlier, but I should have reiterated.

I will look at the update() more closely, as I do have some cases where I suspect updates are not posting, But, I only see those in the context of HA, so it could be something HA is doing, and that's been a moving target in my world since I got back from vacation. I need to test outside of that environment, and should have prior. Sorry for raising an alarm prematurely.

pavoni:
@toggledbits It's a part time task for all of us - very grateful for your help and input.

Whenever you get the chance your input is valued!

Issue #89, opened 2018-04-11
Vera bug in dataversion handling across reloads loses device state


toggledbits:
The polling loop uses long polls, but when Vera restarts Luup, it does not give consistent and accurate device-change data when given a pre-restart value of dataversion. Pull request for a simple patch to clear timestamp and force reload after an exception fixes this.

PR #90, opened 2018-04-11
Additional fix to poll hardening to improve data sync

toggledbits:
As mentioned in issue #89 , Vera gives inaccurate results across Luup reloads if the timestamp parameters (loadtime and dataversion) are used spanning the reload. That is, if a reload occurs, and Vera is then presented with a dataversion from before the reload, it will give inaccurate deltas not just to the initial call, but to subsequent calls as well. Resetting the timestamp on an exception (usually a RequestException resulting from Vera being unreachable during its reload) forces the next long pull to return a full data set, resyncing the local data and providing a stable dataversion with which to continue.

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -243,20 +243,20 @@ class VeraController(object):
         payload.update({
             'id': 'lu_sdata',
         })
-        
+
         LOG.debug("get_changed_devices() requesting payload %s", str(payload))
         r = self.data_request(payload, TIMEOUT*2)
         r.raise_for_status()
 
-        # If the Vera disconnects before writing a full response (as lu_sdata 
-        # will do when interrupted by a Luup reload), the requests module will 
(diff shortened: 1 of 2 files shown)


toggledbits:
By the way, there's really only a one-line change here (plus a comment) in subscribe.py (line 157/8). The remaining changes are my editor removing trailing whitespace from the files.

toggledbits:
Paging @pavoni, and feedback on this one?

pavoni:
Sorry for the long delay - this looks great. 

Thanks for the contribution, will release a new version shortly.

pavoni:
Now released as 0.2.43.

Let me know if you'd like me to do the HA update.

toggledbits:
I'll leave it to you, please. I still need to bring my dev environment up to where I can run tox and handle all of the other steps separate from my live installation.

pavoni:
Added here https://github.com/home-assistant/home-assistant/pull/14540

PR #92, opened 2018-05-21
Add subscription unregister and example usage of subscription with some additional cleanup

jwater7:
Add a listener example, add device accessors by name and id, add device info data accessor, and a subscriber unregister function, rename logger in subscribe as in init so its not confused with a constant


jwater7:
Just a checkpoint that you may want to pick up with some pretty useful changes while I'm looking at what's really going on with the VeraLock and to see if we can improve the performance for them.

PR #101, opened 2018-10-29
Don't consider a curtain as armable

tsvi:
For some reason my vera reports 2 of my covers as armable.
Obviously this is wrong, besides the cover it creates a switch in homeassistant with the same device id.


tsvi:
@pavoni Could you please review this patch and if necessary integrate it into the next HASS release?

It's been very annoying these past couple of months with HASS showing an extra switch for some covers.

pavoni:
Sorry for the delay - now released as `0.2.46`. 

Thanks for the contribution.

PR #105, opened 2018-12-19
Switching dimmer preserves brightness

zintrepid:
Many dimmers preserve their last state. To avoid trashing it, we now
turn on dimmers without explicitly setting the brightness.


pavoni:
This looks to have gone stale. Will close for now - although happy to re-open.

bruring:
Would love to have this implemented. Can it be re-opened?

pavoni:
Hi @bruring. I closed the PR because the developer wasn't able to follow up on @alanfischer 's question and so the PR went state. The feature itself makes sense.

Would be very open to a similar PR in the future, just needs a bit more digging to make sure we don't break anything.

gorbunovav:
@pavoni @bruring Hi! 
As far as I understand the current implementation of the `VeraDimmer` tries to be smarter than Vera itself and uses a custom implementation of the turn on / turn off methods via the brightness level.

This PR just removes that custom implementation, which reverts turn on / turn off methods to the base class - `VeraSwitch`.

Basic `DimmableLight` device specification on the Vera side includes the 'SwitchPower' service, so it supports the same turn on / turn off methods as the basic switch and it is not some vendor-specific feature.

zintrepid:
Original dev here. I cannot say with 100% certainty that this change will not break any switch because I don't own very many switches. I can say that this change will revert back to the Vera's default behavior. The only reason I can see to override the brightness is as a workaround for some vendor that the Vera doesn't work properly with.

As there is no comment as to why the brightness is being overridden, my assumption is that it was an oversight. It appears to me that this library is supposed to wrap the Vera's behavior, so I personally believe that allowing the Vera to implement its own default behavior is appropriate. If there are vendors that are broken, I recommend overriding the brightness only for those vendors. Again, I have no evidence such vendors exist, but I also can't say they don't.

Ultimately it is a judgement call and I can only offer my opinion. I can't really offer much more in terms of compatibility or testing. Thanks for your time!

vangorra:
I have dimmers and regular switches to test against. I'll see what this does.

pavoni:
This is still being requested https://github.com/home-assistant/core/issues/39930 and I think the consequences are pretty small if it does affect other dimmers, so my thinking is to merge this.

Other views?

My plan was to re-open and merge - but I think it's better just to add a new PR - which I'll link here...

PR #110, opened 2019-05-30
 Add support for polling 'alert' data from state data

brandond:
This is the only way to detect some things like code use (which code
was used to unlock the door) and PIN pad lockouts due to excessive
failed attempts.

### DIFF
--- a/examples/device-listener.py
+++ b/examples/device-listener.py
@@ -10,6 +10,7 @@ import pyvera
 
 import time
 
+
 # Parse Arguments
 import argparse
 parser = argparse.ArgumentParser(description='device-listener')
@@ -23,7 +24,8 @@ args = parser.parse_args()
 # Define a callback that runs each time a device changes state
 def device_info_callback(vera_device):
     # Do what we want with the changed device information
-    print('{}_{}: {}'.format(vera_device.name, vera_device.device_id, vera_device.get_all_values()))
+    print('{}_{} values: {}'.format(vera_device.name, vera_device.device_id, vera_device.get_all_values()))
+    print('{}_{} alerts: {}'.format(vera_device.name, vera_device.device_id, vera_device.get_alerts()))
 
 # Start the controller
 controller, _ = pyvera.init_controller(args.url)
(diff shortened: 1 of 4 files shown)


pavoni:
This all looks good to me, although I found this code fragile and hard to test.

I think at the time I had an HA framework that allowed me to run test versions of this code live - but I no longer have that because I'm not as active with HA as I was. Afraid I don't have the time to rebuilt the test environment at the moment.

So am happy to merge if you're confidant that it works and is well tested (and you're willing to sort out issues if we break anything)

jwater7:
@pavoni  If it were me, I'd say merge in https://github.com/pavoni/pyvera/pull/93 and then to minimize regressions we can have new PRs write some unit tests to cover all their new stuff (and use that as a model for all new changes since you're limited now on the project).

pavoni:
@brandond I've merged in the long standing lock PR as suggested by @jwater7, has resulted in a small conflict - could you rebase and resolve? Sorry to have had this in such a low volume repo.

Also let me know your thoughts on testing, and then we can perhaps merge / plan how to release.

brandond:
I agree that this codebase is kinda hard to wrangle. The style is weird and it uses a worker thread where native py3 code would probably go async instead. I was considering running it through a formatter and flake8, but figured I'd try to get the actual thing I wanted in first rather than just show up unannounced with a huge patch.

I'm glad to try to contribute some tests if possible. I have a VeraEdge with two locks on it, so that's all I was able to test locally by actually banging on the API.

brandond:
FWIW, what I was hoping to accomplish here is to be able to send a notification in HA when someone inputs a bad PIN. I discovered that (on my Kwikset locks at least) the PIN failed alert doesn't actually fire until you've failed three times AND it finishes beeping loudly for about 10 seconds.

However, if all you're looking at is the `lu_sdata` api, the `devicedata` field is still empty (because the actual device status hasn't changed - it's still locked) so no callbacks were fired. You have to actually see that the `dataversion` changed and then call the `status` api and parse the alerts to reveal that something happened.

pavoni:
Thanks @brandond, what you've done is well written - you clearly know what you're doing, and great to hear you have managed to test it in a live environment.

Unless @jwater7 disagrees - I'll also merge this and release a new version. I'll find way to check it still work with my lights dimmers etc.

And would be great to have some decent tests - but I think it's quite a big undertaking!

jwater7:
Sounds great!  Nice work

pavoni:
Now releases in `0.3.1`.

PR #112, opened 2019-06-07
Guard against null or missing values

brandond:
After running for a bit, discovered an issue where the 'devices' key was missing from the returned JSON. This wasn't something I'd ran into while testing, but this should prevent exceptions when it happens.


pavoni:
Thanks for catching this. Now released as `0.3.2`.

I'm around next week - but then away for a week. Shall we do the HA PR when I'd back - or do you want to go for it now?

- Greg

PR #115, opened 2019-08-12
Fix for: https://github.com/home-assistant/home-assistant/issues/24987

MarkJenniskens:
because device_id is a string and the keys in _self.devices are integers...

Maybe some other code should be changed so the types match, but this fixes the issue.
Now the device_list is populated (is it really a list, it's always just one device, or am I missing something? )

### DIFF
--- a/pyvera/subscribe.py
+++ b/pyvera/subscribe.py
@@ -95,7 +95,7 @@ class SubscriptionRegistry(object):
 
         for device_id in device_ids:
             try:
-                device_list = self._devices.get(device_id, ())
+                device_list = self._devices.get(int(device_id), ())
                 device_datas = [data for data in device_data_list if data.get('id') == device_id]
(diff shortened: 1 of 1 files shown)


brandond:
It seems like device IDs cannot be counted on to be encoded consistently as ints or strings across all versions of the controller firmware. It would probably be worthwhile to find all the places where device IDs are looked up or compared and ensure that we're reliably casting everything to ints or strings across the board.

MarkJenniskens:
> It seems like device IDs cannot be counted on to be encoded consistently as ints or strings across all versions of the controller firmware, or to even exist at all. It would probably be worthwhile to find all the places where device IDs are looked up or compared and ensure that we're reliably casting everything to ints or strings across the board.

It becomes a problem when using device_id as a key for a dict/list, then the types need to match... (as I understand)

PR #117, opened 2019-09-15
Support for more than 1 controller

vangorra:
Making code run in less of a singleton way by allowing a subscription to use an instance of a controller.
Fixing tests as they were broken.

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -96,7 +96,7 @@ class VeraController(object):
         self.model = None
         self.serial_number = None
         self.device_services_map = None
-        self.subscription_registry = SubscriptionRegistry()
+        self.subscription_registry = SubscriptionRegistry(self)
         self.categories = {}
         self.device_id_map = {}
 
--- a/pyvera/subscribe.py
+++ b/pyvera/subscribe.py
@@ -33,12 +33,13 @@ class PyveraError(Exception):
 class SubscriptionRegistry(object):
     """Class for subscribing to wemo events."""
 
-    def __init__(self):
+    def __init__(self, controller=None):
         """Setup subscription."""
         self._devices = collections.defaultdict(list)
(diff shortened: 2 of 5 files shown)


pavoni:
This looks very sensible to me - any other comments before I merge?

GaryOkie:
Will there be a corresponding update to the docs to clarify what this new subscription feature provides?

And thanks @vangorra for working on this code.   Sure good to see it get some attention, even though I'm not quite sure yet what the update enables.  :)

BTW - has anyone been following discussions at the [Vera developers forum](https://community.getvera.com/t/upcoming-linux-firmware-feedback-needed/209993/4) about a complete rewrite of the Vera controller Linux firmware?    It's long overdue, but having backward compatibility with plugins and API is not planned for the first release, if ever.   Feedback there is welcome, especially as it relates to HA integration and the API.

vangorra:
While I'd prefer to eliminate the singleton altogether, I realize that will break other dependent projects,  alas it must stay. Hence my compromise PR.

I don't think the docs need an update. This change adds expected behavior to the class structure. It's the get_cotroller() call in SubscriptionRegistry that's unusual and seeds confusion when trying to use a different controller.

pavoni:
Now released as `0.3.5`. Thanks for the contribution.

I've just rebuilt my dev machine - and can't currently test integration with Home Assistant - so can someone do this before putting together a PR for HA?

vangorra:
I've confirmed `0.3.5` is working with home assistant.

pavoni:
Fantastic - thanks very much @vangorra 
Are you happy to do the HA PR or would you like me to?

vangorra:
I'll do the PR. I want to remove HA's use of get_controller anyway. That will get us closer to supporting more than one controller in HA.

pavoni:
Perfect - thanks!

PR #118, opened 2019-09-19
Modify parser to handle UserNames that contain space characters.

colohan:
On my Vera some of my UserNames for my locks have both a firstname and lastname listed, with a space as a separator (such as UserName="Jane Smith").  This caused this library to fail to parse them.  Using shlex to split strings handles this case properly since it pays attention to the quoting.


pavoni:
Now released as 0.3.5. Thanks for the contribution.

PR #119, opened 2019-09-27
Add ability for HASS to determine when user enters a valid PIN on lock

colohan:
I really wanted Home Assistant to be able to take an action when I unlock my door with a valid PIN.  But the pyvera library didn't seem to give me the ability to do that in a clean way (and honestly, the Vera HTTP API makes it crazily complex to detect this event....).

So I created a function which makes this information to Home Assistant.  I also added a bunch of comments clarifying what I learned as I read the code.  Please let me know if you'd like anything here changed!

### DIFF
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -182,7 +182,9 @@ class VeraController(object):
     def get_devices(self, category_filter=''):
         """Get list of connected devices.
 
-        category_filter param is an array of strings
+        category_filter param is an array of strings.  If specified, this
+        function will only return devices with category names which match the
+        strings in this filter.
         """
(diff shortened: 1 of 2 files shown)


pavoni:
This looks good to me - and thanks for taking the time to add the very useful comments.
Any thoughts from elsewhere before I merge?

colohan:
FYI -- the corresponding PR for Home Assistant (plumbing this through) is here:  https://github.com/home-assistant/home-assistant/pull/27043

pavoni:
Now released as 0.3.6. Thanks for the contribution.

colohan:
Thanks for making it easy to contribute!  ;-)

PR #120, opened 2019-10-31
Making subscription shutdown more responsive. 

vangorra:
This should fix test timeout issues in https://github.com/home-assistant/home-assistant/pull/28340

Using thread events instead of time.sleep() for the loop to ensure the loop ends almost immediately after calling stop().

### DIFF
--- a/pyvera/subscribe.py
+++ b/pyvera/subscribe.py
@@ -7,7 +7,7 @@ import threading
 import requests
 
 # How long to wait before retrying Vera
-SUBSCRIPTION_RETRY = 10
+SUBSCRIPTION_RETRY = 9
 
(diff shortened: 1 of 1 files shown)


alanfischer:
@pavoni : This looks good to me.  Thoughts?

pavoni:
Sorry for being very slow on this. It certaibly all looks good - I was hoping to find the time to test it out just to double check before merging, but I haven't yet found the time.

@alanfischer if you've already tested it - then I'm very happy to merge and release.

It's certainly a great improvement to my terrible way of managing the event loop thread!

pavoni:
Now released as 0.3.7. Thanks for the contribution - and sorry for the delay.

PR #122, opened 2019-12-05
Adding a lot more functional tests.

vangorra:
Adding a lot more functional tests.
Added code coverage.
Switched test framework to pytest as it works much better.


vangorra:
Any progress on reviewing this PR? I have another change that adds mac, manufacture and model to the VeraDevice but it depends on the test in this PR.

pavoni:
Sorry for the slow reply. My general reaction is it's great to have these!

I've added you as a contributer.

Haven't been through the tests in detail, but this is a big step in the right direction IMHO. Only suggestion is that I think we should add a `requirements_test.txt` file.

This is what worked for me
```
flake8>=3.6.0
pylint>=2.2.2
pytest>=4.0.2
pytest-cov>=2.8.1
mock>=3.0.5
```

vangorra:
I'm glad you brought up the requirements file. I held off implementing something like that as I didn't want to overstep my bounds.
Since you are okay with me as a contributor, I'll add pyproject.toml file, as that is Python's new way of doing this. Additionally, I'll add CI configs too. Basically, I'll be copying a lot of the devops code from https://github.com/vangorra/python_withings_api

PR #150, opened 2021-09-10
Merge pull request #148 from pavoni/handle_comms_failure_better

sresam89:
Since the code has been using the following line (getting the hash of the password as well) python was not able to handle the character length more than 254 (which it is usually when doorcode is programmed)

Original code on init.py line 1210
`_, _, pin, name = code_addrs[2:]`

error without the fix
_Problem parsing pin code string 1,1,2021-08-12 16:35:38,2021-09-01 18:52:22,****,Member,2E65312E07791C05095B;: too many values to unpack (expected 4)_

Suggested code: (to remove the hash, thereby eliminating the error of character length)
`_, _, pin, name = code_addrs[2:6]`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pyvera"
-version = "0.3.12"
+version = "0.3.13"
 description = "Python API for talking to Veracontrollers"
 
 license = "GPL2"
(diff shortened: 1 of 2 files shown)


sresam89:
Closing this since starting another with a fork

PR #151, opened 2021-09-10
Update __init__.py

sresam89:
Fixing the "Problem parsing pin code string too many values to unpack (expected 4) "


sresam89:
@pavoni or @vangorra  could you please approve the pull request

README.md
```diff
@@ -50,6 +50,11 @@ If you have locks - this will show you information about them.
 $ ./examples/show_lock_info.py -u http://192.168.1.161:3480/
 ~~~~
 
+Set a new door lock code
+~~~~
+$ ./examples/set_and_delete_door_code.py -u http://192.168.1.161:3480/ -n "John Doe" -p "12345678"
```
pavoni, line 55: The readme refers to `set_and_delete_door_code` - but the PR contains `delete_door_code` and `set_door_code`

examples/delete_door_code.py
```diff
@@ -0,0 +1,56 @@
...
+                # You can only delete the slots on the lock
+                # You should also know which slot number you want to delete as an integer
+                # You can lookup name and slots using device.get_pin_codes() api
+                # This example deletes a previously added slot by getting the no:of slots already allocated                
+                result = device.clear_slot_pin(slot=args.slot)
+                if result.status_code==200:
+                    print("\nCommand succesfully sent to Lock \
+                    \nWait for the lock to process the request")
+                else:
+                    print("\nLock command "+result.text)
```
pavoni, line 48: Perhaps better "\nLock command failed "

examples/set_door_code.py
```diff
@@ -0,0 +1,61 @@
...
+    )
+    args = parser.parse_args()
+
+    # Start the controller
+    controller = VeraController(args.url)
+    controller.start()
+
+    try:
+        # Get a list of all the devices on the vera controller
+        all_devices = controller.get_devices()
```
pavoni, line 36: `get_devices` same comment as above

examples/delete_door_code.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+    args = parser.parse_args()
+
+    # Start the controller
+    controller = VeraController(args.url)
+    controller.start()
+
+    try:
+        # Get a list of all the devices on the vera controller
+        all_devices = controller.get_devices()
```
pavoni, line 33: `get_devices` takes a second parameter to filter the device type - which might be slightly more elegant.

pyvera/__init__.py
```diff
@@ -545,25 +545,26 @@ def set_service_value(
         self,
         service_id: Union[str, Tuple[str, ...]],
         set_name: str,
-        parameter_name: str,
-        value: Any,
+        parameter: dict,
```
pavoni, line 548: I no longer have a live vera hub - which limits my ability to check changes. As such while I can see why you want to change the parameter structure -but I'm a little reluctant because testing the refactoring needed.

Even though it's less elegant I'd be happier for you to just add an extra call for the lock api - and leave the "set_service_value" method unchanged.

pyvera/__init__.py
```diff
@@ -545,25 +545,26 @@ def set_service_value(
...
     ) -> None:
         """Set a variable on the vera device.
 
         This will call the Vera api to change device state.
         """
         payload = {
             "id": "lu_action",
-            "action": "Set" + set_name,
+            "action": set_name,
```
pavoni, line 556: See above

pyvera/__init__.py
```diff
@@ -545,25 +545,26 @@ def set_service_value(
...
             "serviceId": service_id,
-            parameter_name: value,
         }
+        for param in parameter:
```
pavoni, line 559: See above

pyvera/__init__.py
```diff
@@ -803,7 +804,7 @@ class VeraSwitch(VeraDevice):
 
     def set_switch_state(self, state: int) -> None:
         """Set the switch state, also update local state."""
-        self.set_service_value(self.switch_service, "Target", "newTargetValue", state)
+        self.set_service_value(self.switch_service, "SetTarget", {"newTargetValue", state})
```
pavoni, line 807: See above

pyvera/__init__.py
```diff
@@ -856,7 +857,7 @@ def set_brightness(self, brightness: int) -> None:
             percent = round(brightness / 2.55)
 
         self.set_service_value(
-            self.dimmer_service, "LoadLevelTarget", "newLoadlevelTarget", percent
+            self.dimmer_service, "SetLoadLevelTarget", "newLoadlevelTarget", percent
```
pavoni, line 860: See above

pyvera/__init__.py
```diff
@@ -904,7 +905,7 @@ def set_color(self, rgb: List[int]) -> None:
 
         target = ",".join([str(c) for c in rgb])
         self.set_service_value(
-            self.color_service, "ColorRGB", "newColorRGBTarget", target
+            self.color_service, "SetColorRGB", "newColorRGBTarget", target
```
pavoni, line 908: See above

pyvera/__init__.py
```diff
@@ -934,7 +935,7 @@ class VeraArmableDevice(VeraSwitch):
     def set_armed_state(self, state: int) -> None:
         """Set the armed state, also update local state."""
         self.set_service_value(
-            self.security_sensor_service, "Armed", "newArmedValue", state
+            self.security_sensor_service, "SetArmed", "newArmedValue", state
```
pavoni, line 938: See above

pyvera/__init__.py
```diff
@@ -1020,7 +1021,7 @@ def set_level(self, level: int) -> None:
         Scale is 0-100
         """
         self.set_service_value(
-            self.dimmer_service, "LoadLevelTarget", "newLoadlevelTarget", level
+            self.dimmer_service, "SetLoadLevelTarget", "newLoadlevelTarget", level
```
pavoni, line 1024: See above

pyvera/__init__.py
```diff
@@ -1035,10 +1036,18 @@ class VeraLock(VeraDevice):
 
     def set_lock_state(self, state: int) -> None:
         """Set the lock state, also update local state."""
-        self.set_service_value(self.lock_service, "Target", "newTargetValue", state)
+        self.set_service_value(self.lock_service, "SetTarget", {"newTargetValue": state})
```
pavoni, line 1039: See above

pyvera/__init__.py
```diff
@@ -1229,7 +1238,7 @@ def set_temperature(self, temp: float) -> None:
         """Set current goal temperature / setpoint."""
 
         self.set_service_value(
-            self.thermostat_setpoint, "CurrentSetpoint", "NewCurrentSetpoint", temp
+            self.thermostat_setpoint, "SetCurrentSetpoint", "NewCurrentSetpoint", temp
```
pavoni, line 1241: See above

pyvera/__init__.py
```diff
@@ -1255,7 +1264,7 @@ def get_current_temperature(self, refresh: bool = False) -> Optional[float]:
     def set_hvac_mode(self, mode: str) -> None:
         """Set the hvac mode."""
         self.set_service_value(
-            self.thermostat_operating_service, "ModeTarget", "NewModeTarget", mode
+            self.thermostat_operating_service, "SetModeTarget", "NewModeTarget", mode
```
pavoni, line 1267: See above

pyvera/__init__.py
```diff
@@ -1283,7 +1292,7 @@ def turn_auto_on(self) -> None:
 
     def set_fan_mode(self, mode: str) -> None:
         """Set the fan mode."""
-        self.set_service_value(self.thermostat_fan_service, "Mode", "NewMode", mode)
+        self.set_service_value(self.thermostat_fan_service, "SetMode", {"NewMode", mode})
```
pavoni, line 1295: See above

pyvera/__init__.py
```diff
@@ -1195,7 +1204,7 @@ def get_pin_codes(self, refresh: bool = False) -> List[LockCode]:
             LOG.error("Got unsupported string %s: %s", val, ex)
 
         # Loop to create a list of codes
-        codes = []
+        codes = {}
```
pavoni, line 1207: If slots are integers - why do you want to change this to a dictionary?

pavoni (changes requested):
Thanks for submitting this.

I no longer have a live vera hub myself. I still have it - but even if I fire it up it no longer has many devices attached - so it's not a great test.

For that reason I'd like to keep changes as simple as possible to minimise the risk for people who depend on this code.

As such I've suggested leaving the parameters of `set_service_value` unchanged - and just adding another  similar method to do what you need for the lock.

Not my usual preference - but in this case I'd prefer to limit change to the lock code.

pyvera/__init__.py
sresam89: I did not request this change, it should have been from the tag I forked from
I made chagne to the line 1219
`_, _, pin, name = code_addrs[2:6]`  added `:6` to the list index

examples/delete_door_code.py
sresam89: I was following the existing code flow, minimizing no:of breaking changes

README.md
sresam89: I will update the readme on this

examples/delete_door_code.py
sresam89: I believe it would be  better to show the user what was the reason code is rather throwing a general message!!
There could be several reasons
1) Bad slot number
2) Slot already empty and nothing to clear
3) Lock is busy
4) Connection error (request timeout....)

Your thoughts?

examples/set_door_code.py
sresam89: I was following the existing code flow, minimizing no:of breaking changes

pyvera/__init__.py
sresam89: I can do that....

examples/delete_door_code.py
pavoni: I was just suggesting you change the text so it was more obvious it was an error. Happy with including rest of detail as you have.

examples/set_door_code.py
pavoni: My suggestion would improve it slightly and avoid the need to do the `isinstance` check.

pyvera/__init__.py
pavoni: I guess you may need to rebase. 

But you’re branch currently includes this change which isn’t in the library.

pyvera/__init__.py
pavoni: Open minded.  Perhaps `call_service_with_params` or something like that?

examples/set_door_code.py
sresam89: I could, but I would have to change the get_devices() method on all over the code, which could be a breaking change on hassio

pavoni: Take a look at the code, there is an optional second parameter. You only need to change your call!

https://github.com/pavoni/pyvera/blob/941823e1a13c8ea2d4dd398287939bf0c6f970f6/pyvera/__init__.py#L196

pyvera/__init__.py
sresam89: used set_door_code_values

pyvera/__init__.py
sresam89: Addressed

pyvera/__init__.py
sresam89: Addressed

pyvera/__init__.py
sresam89: Addressed

pyvera/__init__.py
sresam89: Addressed

pyvera/__init__.py
sresam89: Sorry My bad, I did change code to be dict over list. 
I am formatting this more elegantly, over output string being just a CSV list.

Previously it was
` codes.append((slot, name, pin))` I changed to be `codes[slot]={'name': name, 'pin': pin}`

There is no dependency on hassio for this, this will be a new feature request there

pavoni: You changed codes from an array to a dict.

please put it back.

sresam89: Is the Change from array to dict not acceptable? Is it breaking anything else?

pavoni:
Just for clarity - you also need to fix the build error. 

Looks like your imports are not sorted correctly.

pyvera/__init__.py
pavoni: You already know my views.

sresam89: Sorry not getting emotions over comments.
I would like to hear it if it is not acceptable and understand the reasoning why
I would love to discuss/hear about the same on other chat/messenger/meeting platforms if necessary

pavoni: As you already know I’m trying to keep this code stable.

I don’t currently see any good reason for changing `codes` from an array to a dict.

sresam89: Having the output as dict helps a ton on the hassio side or other applications that will have to parse a CSV otherwise. 
Looks like that is not good enough of an argument for you

pavoni: It doesn’t make any sense to me.

As far as I can see you’re creating a dictionary with a key of `slot` which is an integer.

Each value of the `codes` dictionary is itself a dictionary - and I agree these make sense because the keys are ‘ name` and `pin`.

I expect if you just change back the initial assignment of `codes` you code will work because you’re assigning into it with an array assignment anyway (I agree the assignment is better than the append)

So honestly it just looks like an mistake.

sresam89: Hope you agree to correct this mistake someday. 

Probably another PR to fight for. 

Am reverting to original for now.

sresam89:
Made all the changes requested (except for passing the deviceIsntance to lock) and pushed the code, let me know your comments

> Just for clarity - you also need to fix the build error.
> 
> Looks like your imports are not sorted correctly.

Could you help me with this
Am running this on a windows machine. Not able to use your build scripts

pyvera/__init__.py
pavoni: It’s not a mistake on my part.

Another comment like this I will stop reviewing and close this PR.

examples/delete_door_code.py
pavoni: I think you still need to make this change.

pyvera/__init__.py
```diff
@@ -549,7 +549,6 @@ def set_service_value(
         value: Any,
     ) -> None:
         """Set a variable on the vera device.
-
```
pavoni, line 552: Please put back this blank line to maintain consistency with the rest of the code base.

sresam89: Done

examples/set_door_code.py
sresam89: Done

examples/delete_door_code.py
sresam89: Done

pyvera/__init__.py
sresam89: addressed

pavoni:
What’s currently failing is ‘black’. Even if you can’t run the build script, you should be able to run the individual utilities.

You can see what’s being run here: https://github.com/pavoni/pyvera/blob/master/scripts/build.sh

If you can’t track down the issues I’ll check out your branch and run them, but I won’t be able to do that until mid next week.

pavoni:
You can read the build results here (although the black error isn’t very helpful!)

https://github.com/pavoni/pyvera/pull/151/checks?check_run_id=3615579670

sresam89:
> What’s currently failing is `black`. Even if you can’t run the build script, you should be able to run the individual utilities.
> 
> You can see what’s being run here: https://github.com/pavoni/pyvera/blob/master/scripts/build.sh
> 
> If you can’t track down the issues I’ll check out your branch and run them, but I won’t be able to do that until mid next week.

Fixed black error wish

pavoni:
You’ve still got black error on your changes to `__init__.py`

sresam89:
> You’ve still got black errors on your changes to `__init__.py`

Not on my local machine @pavoni see the screenshot
image

Commit Id 0a69a9648af178d0e2d6e5a7b333bbee5eeffd97 checkout clean with black on my local machine

pavoni:
Don’t know if this helps you get stuff running. https://dev.to/bowmanjd/getting-started-with-python-poetry-3ica

Perhaps double check which black version you have.

Otherwise I’ll run it locally next week and let you know the problem.

Also worth making sure flake8, pylint etc run cleanly.

sresam89:
> Don’t know if this helps you get stuff running. https://dev.to/bowmanjd/getting-started-with-python-poetry-3ica
> 
> Perhaps double check which black version you have.
> 
> Otherwise I’ll run it locally next week and let you know the problem.
> 
> Also worth making sure flake8, pylint etc run cleanly.

i got hold a linux pc and ran the build script, and now lint complaint about the following, which does not makes sense to me.
Strangely the same error goes away if include the `isinstance` check you asked me to remove
image

sresam89:
Adding a test commit to verify possible Lint Issues 
[https://github.com/python/mypy/issues/8763](url)

sresam89 (commented):
Adding test commit to verify lint issues

sresam89:
I was able to fix almost every error please review the changes, also added test cases

pavoni:
The build is still failing saying 4 files would be reformatted.

I’ll run locally and post the results in the next few days if you’re unable to reproduce with the ‘build’ script locally.

sresam89 (commented):
I was able to resolve almost every build error, and also added test cases please review

pavoni:
You can see the failing build here!

https://github.com/pavoni/pyvera/pull/151/checks?check_run_id=3646361399

sresam89 (commented):
Black shows only 3 files for me. updated them

pavoni:
This is the change black made for me:-

```
diff --git a/pyvera/__init__.py b/pyvera/__init__.py
index dffc8bc..4e0a4f6 100755
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -566,20 +566,13 @@ class VeraDevice:
         )

     def set_door_code_values(
-        self,
-        service_id: Union[str, Tuple[str, ...]],
-        operation: str,
-        parameter: dict,
+        self, service_id: Union[str, Tuple[str, ...]], operation: str, parameter: dict
     ) -> requests.Response:
         """Add or remove door code on the vera Lock.

         This will call the Vera api to change Lock code.
         """
-        payload = {
-            "id": "lu_action",
-            "action": operation,
-            "serviceId": service_id,
-        }
+        payload = {"id": "lu_action", "action": operation, "serviceId": service_id}
         for param in parameter:
             payload[param] = parameter[param]
         result = self.vera_request(**payload)
(END)
```

sresam89:
> diff --git a/pyvera/__init__.py b/pyvera/__init__.py

Which version of black are you using?

sresam89 (commented):
I think i figured the problem.
I was using the 21.9b0 and you were using some version of 19.x

pavoni:
‘poetry’ should be taking care of that. You can see ‘19.3b0’ in the lock file.

if you’re comfortable - I’ll just try and commit an update of ‘__init__’ that works to your branch.

pavoni:
Black still wants you to remove the final `,`

```
diff --git a/pyvera/__init__.py b/pyvera/__init__.py
index 648dbff..4e0a4f6 100755
--- a/pyvera/__init__.py
+++ b/pyvera/__init__.py
@@ -566,7 +566,7 @@ class VeraDevice:
         )

     def set_door_code_values(
-        self, service_id: Union[str, Tuple[str, ...]], operation: str, parameter: dict,
+        self, service_id: Union[str, Tuple[str, ...]], operation: str, parameter: dict
     ) -> requests.Response:
         """Add or remove door code on the vera Lock.

(END)
```

pavoni (approved):

pavoni:
Apart from the final black change I think this is now GTG.

Happy to merge and release when the build is clean.

sresam89:
> ‘poetry’ should be taking care of that. You can see ‘19.3b0’ in the lock file.
> 
> if you’re comfortable - I’ll just try and commit an update of ‘**init**’ that works to your branch.

poetry is not doing that with respect to black (I had to install it manually)
```
(.venv) sreeram@ubuntu-sam:~/hassio/pyvera$ poetry install
Installing dependencies from lock file                                 

No dependencies to install or update

Installing the current project: pyvera (0.3.14)
(.venv) sreeram@ubuntu-sam:~/hassio/pyvera$ pip list | grep b*
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.

(.venv) sreeram@ubuntu-sam:~/hassio/pyvera$ which black
/home/sreeram/.local/bin/black
(.venv) sreeram@ubuntu-sam:~/hassio/pyvera$ black --version
black, version 19.10b0
```

I uninstalled black at this point to show you
```
(.venv) sreeram@ubuntu-sam:~/hassio/pyvera$ poetry install
Installing dependencies from lock file

No dependencies to install or update

Installing the current project: pyvera (0.3.14)
(.venv) sreeram@ubuntu-sam:~/hassio/pyvera$ black --version
bash: /home/sreeram/.local/bin/black: No such file or directory
(.venv) sreeram@ubuntu-sam:~/hassio/pyvera$ 
```

sresam89 (commented):
updating black changes (hopefully the last time)
you can make anymore changes, am ok with that

sresam89:
How long does this take to update on https://pypi.org/project/pyvera/ ??

pavoni:
I need to do a release - which I’ll do later today.

The update is then almost instant.

pavoni:
Now released as 0.3.14.

Thanks for the contribution.

sresam89:
I know we had a conversation on this already. Is it worth re-visiting the "pincode" being dict versus list ?
I totally understand your view on maintaining the stable code base, just trying an attempt to make it better.

pavoni:
For items which are indexed only by an integer - the best data structure is really an array rather than a dict.

But unless there are a lot of pins - and a decent amount of looking up going on I think the advantages of changing to another data structure are small - and offset by the fact that we might break existing code using the API.

The library is't being actively developed - so I think t's wise to keep non critical changes to a minimum.

PR #152, opened 2022-02-06
Fix VeraLock#set_new_pin

jnewland:
This PR updates the parameter used in the `SetPin` Vera call to use the `newPin` parameter, which matches what I see [documented](http://wiki.micasaverde.com/index.php/Luup_UPnP_Variables_and_Actions#DoorLock1) and what I am able to observe by making a similar change in the Vera UI and observing traffic via the web inspector.

Before this change, I received an error when running the included `set_door_code.py` example:

```
Lock command ERROR: The PIN entered is too long or short. Please enter a PIN of the specified length.
```

After this change I'm able to successfully set a PIN code on one of my locks.

In support of testing this change, I updated the included examples that set/delete PIN codes to require a device ID instead of repeating the same command to all locks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,14 +50,19 @@ If you have locks - this will show you information about them.
 $ ./examples/show_lock_info.py -u http://192.168.1.161:3480/
 ~~~~
 
-Set a new door lock code
+View existing locks and PINs:
 ~~~~
-$ ./examples/set_door_code.py -u http://192.168.1.161:3480/ -n "John Doe" -p "12345678"
+$ ./examples/show_lock_info.py -u http://192.168.1.161:3480/
 ~~~~
 
-Clear a existing door lock code
+Set a new door lock code on device 335:
 ~~~~
-$ ./examples/delete_door_code.py -u http://192.168.1.161:3480/ -n "John Doe" -p "12345678"
+$ ./examples/set_door_code.py -u http://192.168.1.161:3480/ -i 335 -n "John Doe" -p "5678"
+~~~~
+
+Clear a existing door lock code from device 335:
+~~~~
+$ ./examples/delete_door_code.py -u http://192.168.1.161:3480/ -i 335 -n "John Doe"
 ~~~~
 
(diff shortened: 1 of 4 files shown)


jnewland:
@pavoni gentle bump for a review on this, motivated by a desire to revisit https://github.com/home-assistant/core/pull/60310 and push it across the finish line. I see in other PRs you no longer have a live hub: I'm happy to do any sort of testing you'd like in support of review. Thanks in advance!

pavoni:
Really sorry - this got lost in my email!

Will take a proper look now.

Also let me know if you're interested in taking on the repo - having moved away from using a vera hub - I don't feel I'm in a good position to be the main maintainer!

pavoni:
This looks good to me.

Assume you'd like me to tag and release?

pavoni:
@jnewland Now released as 0.3.15. Given the notes below please test carefully!

I'd previously updated my local python version - so I had some problems building locally - and had to disable `mypi`.

A lot of the libraries are out of date - and since the main use of the library is with Home Assistant it might make sense to drop support for older versions of python - and just focus on 3.9.

I added an issue here: https://github.com/pavoni/pyvera/issues/154

This will take some work though as there are quite a few lint errors!

Thanks for joining as a contributor, very happy to help with releases etc if you're open to do some of the dev / upgrade work.

jnewland:
Awesome, I'll test 0.3.15 out with my local HA install later today to confirm everything is working. 

Totally agree about focusing on 3.9, I'd be happy to take a shot at that soon. Will leave more notes in that other issue once I look more deeply. ⚡

sresam89:
@jnewland  thanks for updating the code.
I had only one lock to test with the time created the PR, and it managed to work.
Since this was the first attempt to get this feature, I have not seen many using it in hassio.

Thanks again

jnewland:
@sresam89 unfortunately I'm no longer using Vera and thus don't plan to follow up on this any further.

PR #165, opened 2024-10-25
Use SPDX license identifier

cdce8p:
Followup to #164

Poetry recommends the use of the official SPDX license identifier. For `GPL-2.0-only` that will add the license classifier automatically.


cdce8p:
This PR has been open without a response for some time now. Going to close it. Please feel free to reuse any part of it if you like.

